PR #61, opened 2015-04-20
Error (Yiddish)

Liongrass:
I very much assume "Yiddish (German)" is supposed to mean "Yiddish (Germany)", as Yiddish and German are two separate languages and given the DE country code in yi-DE .


naconner:
@Liongrass Thanks for the correction, Leonhard.

PR #62, opened 2015-06-29
Fix parsing for North Carolina licenses

naconner:
Awaiting field testing.


naconner:
Already committed in master.

PR #63, opened 2015-06-29
Add AUD and JPY bill lengths

naconner:
Awaiting field testing.


naconner:
Already committed in master.

PR #71, opened 2016-04-10
Restart camera after periodic reboot

naconner:
The bash script fixes the issue noted on Aaeons where the camera may not scan and which is not solved by l-m restarts. Currently, this is resolved by a full machine reboot or by unplugging and replugging the camera's USB cable.

If whatever is causing the non-scanning doesn't happen in under three hours of being reconnected, this shouldn't recur.

Tested well on an Aaeon. Still need to test compatibility on Nexus, though this has been much less of an issue on them.

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -153,7 +153,6 @@ Brain.prototype.run = function run () {
   var callback = function () {
     self._transitionState('restart')
     console.log('Scheduled restart after idle time.')
-    var cameraRestart = cp.exec('bash /opt/apps/machine/lamassu-machine/lib/camera_restart.sh')
     process.exit()
   }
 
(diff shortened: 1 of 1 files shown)


lib/brain.js
```diff
@@ -153,7 +153,7 @@ Brain.prototype.run = function run () {
   var callback = function () {
     self._transitionState('restart')
     console.log('Scheduled restart after idle time.')
-    var cameraRestart = cp.exec('bash /opt/apps/machine/lamassu-machine/lib/util/camera_restart.sh')
+    var cameraRestart = cp.exec('bash /opt/apps/machine/lamassu-machine/lib/camera_restart.sh')
```
naconner: Should be a relative path anyway. Testing.

joshmh:
Nice! It probably won't work on Nexus because I don't think the grep command is installed. But there should be a workaround, or we could just disable for Nexus.

naconner:
@joshmh Looks like this needs to be run after l-m starts, not before. Know of a place in brain.js to insert cameraRestart for this?

naconner:
@joshmh We could update busybox on the Nexus. I had been using the latest versions on mine a little while back without issue: http://busybox.net/downloads/binaries/latest/busybox-armv7l

joshmh:
Is this still relevant?

naconner:
The camera drop issue can still present in 0.5.20-13. However, I have not tested this code in a while.

joshmh:
Closing for now.

PR #77, opened 2016-08-28
add MYR, JMD bill lengths

naconner:


joshmh:
Is this still relevant?

naconner:
Yes, these aren't in v3 yet: https://github.com/lamassu/lamassu-machine/blob/v3/lib/puloon/puloon_data.js

PR #85, opened 2017-08-21
Add SMS verification screen 

goga-m:
This commit contains the following bug: When a user returns from sms verification screen and inserts bills, the following error is thrown in l-m and exits

LOG AssertionError [ERR_ASSERTION]: bill fiat is positive, can't start tx


lib/brain.js
```diff
@@ -1368,6 +1378,15 @@ Brain.prototype._billRead = function _billRead (data) {
 
     return
   }
+  // TODO: Check if threshold is 0 and popup the sms screen without bill insertion
+  if(this.trader.smsVerificationActive && fiat.gte(this.trader.smsVerificationThreshold)) {
```
joshmh: Looks like there's no code to reject the bill, which is part of the spec. The bill should be rejected, which will clear ``this.bill``, which should prevent the error you were seeing when subsequently inserting bills.

lib/brain.js
```diff
@@ -900,18 +906,22 @@ Brain.prototype.smsCompliance = function smsCompliance (nextScreen) {
   flow.on('success', () => {
     const phone = flow.phone
     this.flow = null
-
     this.tx = Tx.update(this.tx, {phone})
-    this.startScreen(true)
+
+    if(returnState) {
```
joshmh: Lots of flows are not handled correctly. For instance, if you enter in the wrong security code 3 times, you will be sent back to the start screen instead of being given a chance to send your coins. These flows have to modified from the original code which was only designed to work with a threshold of 0.

PR #99, opened 2018-01-30
Fix for bin/sencha-brain - set execution environment

sethbergman:
Added to top of file:
`#!/usr/bin/env node`


joshmh:
``sencha-brain`` hasn't been used in a long time, so it should be deleted.

PR #104, opened 2018-02-28
Resync machine time to prevent clock skew error

naconner:
Upon each periodic reboot or machine start, this forces the machine clock to resync, avoiding `clock skew with lamassu-machine too high` errors.


naconner:
Needs to not trigger on Nexus, or have a Nexus-specific script. Looking into that.

joshmh:
Committed in a separate commit.

PR #107, opened 2018-03-14
Extract buy/sell to i18n

buildog:
Currently the buy/sell label is not translated, I let you generate the i18n base and apply it to crowdin.


joshmh:
Implemented here: https://github.com/lamassu/lamassu-machine/commit/4741d70e17f83431f12afb2eece78812e6433d5d

PR #114, opened 2018-05-29
Terms screen added

lidio601:
Relies on https://github.com/lamassu/lamassu-server/pull/118

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -756,6 +756,10 @@ Brain.prototype._processReal = function _processReal (req) {
     case 'blockedCustomerOk':
       this._idle()
       break
+    case 'termsAccepted':
+      this.acceptTerms()
+      this._start()
+      break
   }
 }
 
@@ -1468,6 +1472,14 @@ Brain.prototype.isCompliant = function isCompliant (amount) {
   return _.isEmpty(this.nonCompliantTiers(amount))
 }
 
+Brain.prototype.acceptTerms = function () {
+  const updateRec = {
+    termsAccepted: true
+  }
+  const update = _.assignAll([this.tx, updateRec])
+  this.tx = Tx.update(this.tx, update)
+}
+
 Brain.prototype._start = function _start () {
   if (this.startDisabled) return
   if (this.isLowBalance()) return this._timedState('balanceLow')
@@ -1481,6 +1493,13 @@ Brain.prototype._start = function _start () {
     minimumTx: this.billValidator.lowestBill(coin.minimumTx),
     cryptoNetwork: coin.cryptoNetwork
   }
+
+  if (!this.tx.termsAccepted && !this.trader.terms.active) {
+    // terms are automatically accepted
+    // if terms screen is not enabled
+    updateRec.termsAccepted = true
+  }
+
   const update = _.assignAll([this.tx, updateRec])
   this.tx = Tx.update(this.tx, update)
 
@@ -1491,7 +1510,13 @@ Brain.prototype._start = function _start () {
     return this.smsCompliance()
   }
 
-  this._startAddressScan()
+  // check if terms screen is enabled
+  if (!this.tx.termsAccepted) {
+    this._showTermsScreen()
+  } else {
+    this._startAddressScan()
+  }
+
   this.browser().send({tx: this.tx})
 }
 
@@ -1507,6 +1532,16 @@ Brain.prototype._cancelIdScan = function _cancelIdScan () {
   this.startDisabled = true
   this._bye()
   this.scanner.cancel()
+}
+
+Brain.prototype._showTermsScreen = function _startAddressScan () {
+  this._transitionState('termsScreen')
+
+  // after 60 sec return to main screen
+  this.screenTimeout = setTimeout(() => {
+    if (this.state !== 'termsScreen') return
+    this._idle() // TODO: check why is not showing start screen
+  }, this.config.qrTimeout)
 }
 
 Brain.prototype._startAddressScan = function _startAddressScan () {
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -605,7 +605,8 @@ Brain.prototype._connectedBrowser = function _connectedBrowser () {
     coins: this.trader.coins,
     twoWayMode: this.twoWayMode(),
     wifiList: wifiList,
-    rates
+    rates,
+    terms: this.trader.terms
   }
 
   this.browser().send(fullRec)
@@ -757,8 +758,8 @@ Brain.prototype._processReal = function _processReal (req) {
       this._idle()
       break
     case 'termsAccepted':
-      this.acceptTerms()
-      this._start()
+      this.fastUpdateTx({termsAccepted: true})
+      this.startScreen()
       break
   }
 }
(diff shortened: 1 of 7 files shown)


lib/brain.js
```diff
@@ -760,6 +760,10 @@ Brain.prototype._processReal = function _processReal (req) {
       return this.lightningMode(true)
     case 'lightningNo':
       return this.lightningMode(false)
+    case 'termsAccepted':
+      this.acceptTerms()
+      this._start()
```
joshmh: Need to handle cash-out, as well.

lib/brain.js
```diff
@@ -1472,6 +1476,14 @@ Brain.prototype.isCompliant = function isCompliant (amount) {
   return _.isEmpty(this.nonCompliantTiers(amount))
 }
 
+Brain.prototype.acceptTerms = function () {
+  const updateRec = {
+    termsAccepted: true
+  }
+  const update = _.assignAll([this.tx, updateRec])
+  this.tx = Tx.update(this.tx, update)
```
joshmh: Should be:

```
this.fastUpdateTx({termsAccepted: true})
```

lib/brain.js
```diff
@@ -1485,6 +1497,13 @@ Brain.prototype._start = function _start () {
     minimumTx: this.billValidator.lowestBill(coin.minimumTx),
     cryptoNetwork: coin.cryptoNetwork
   }
+
+  if (!this.tx.termsAccepted && !this.trader.terms.active) {
+    // terms are automatically accepted
+    // if terms screen is not enabled
+    updateRec.termsAccepted = true
+  }
+
```
joshmh: We should not record that user accepted terms if he didn't, even if there are no terms. We should handle this case by checking whether terms are enabled to determine whether to show the screen. I would remove this section.

lib/brain.js
```diff
@@ -1495,7 +1514,13 @@ Brain.prototype._start = function _start () {
     return this.smsCompliance()
   }
 
-  this._startAddressScan()
+  // check if terms screen is enabled
+  if (!this.tx.termsAccepted) {
+    this._showTermsScreen()
+  } else {
+    this._startAddressScan()
```
joshmh: What about cash-out?

lib/brain.js
```diff
@@ -1513,6 +1538,16 @@ Brain.prototype._cancelIdScan = function _cancelIdScan () {
...
+Brain.prototype._showTermsScreen = function _startAddressScan () {
+  this._transitionState('termsScreen')
+
+  // after 60 sec return to main screen
+  this.screenTimeout = setTimeout(() => {
+    if (this.state !== 'termsScreen') return
+    this._idle() // TODO: check why is not showing start screen
+  }, this.config.qrTimeout)
+}
+
```
joshmh: This is unnecessary, just use:

```
this._timedState('termsScreen')
```

lib/trader.js
```diff
@@ -40,6 +40,8 @@ const Trader = function (protocol, clientCert, connectionInfo) {
   this.pollRetries = 0
   this.state = {state: 'initial', isIdle: false}
   this.configVersion = null
+  this.terms = {active: false}
```
joshmh: Wouldn't this be simpler as a boolean and not an object?

ui/src/app.js
```diff
@@ -232,6 +232,9 @@ function processData (data) {
     case 'blockedCustomer':
       blockedCustomer()
       break
+    case 'termsScreen':
+      showTermsScreen()
+      break
```
joshmh: This is unnecessary since it's handled by the default case.

ui/src/app.js
```diff
@@ -608,6 +618,38 @@ function setWifiList (recs, requestedPage) {
...
+    if (!elem.data('default')) elem.data('default', elem.html())
+    label = label || elem.data('default')
+    // replace **text** with <b>text</b>
+    let matches = label.match(/(\*\*.[^*]*\*\*)/g);
+    matches && matches.forEach(function (match) {
+      label = label.replace(match, match
+        .replace('**', '<b>')
+        .replace('**', '</b>'))
+    })
+    elem.html(label)
```
joshmh: This should be done on the server side, with a markdown library.

ui/start.html
```diff
@@ -1319,6 +1320,21 @@ <h1 class="js-i18n">Out of cash</h1>
         </section>
       </section>
 
+      <section class="viewport terms_screen_state js-terms-screen" data-tr-section="terms">
+        <section class="info">
+          <h1 class="js-i18n js-terms-title">Terms and Conditions</h1>
```
joshmh: No need to supply defaults. We wouldn't want to show these defaults anyway.

ui/start.html
```diff
@@ -1219,6 +1219,7 @@ <h1 class="js-i18n">Network Down</h1>
           <section class="description">
             <p class="js-i18n">Sorry, we're down for the moment. Please come back later.</p>
           </section>
+          <section class=actions">
```
joshmh: Missing "

ui/src/app.js
```diff
@@ -115,6 +114,7 @@ function processData (data) {
   if (data.readingBill) readingBill(data.readingBill)
   if (data.cryptoCode) translateCoin(data.cryptoCode)
   if (data.tx && data.tx.cashInFee) setFixedFee(data.tx.cashInFee)
+  if (data.terms) setTermsScreen(data.terms)
```
joshmh: `data.terms` doesn't seem to ever be sent from `brain.js`.

lib/brain.js
```diff
@@ -1513,6 +1538,16 @@ Brain.prototype._cancelIdScan = function _cancelIdScan () {
   this.scanner.cancel()
 }
 
+Brain.prototype._showTermsScreen = function _startAddressScan () {
```
joshmh: Don't forget to change the function name, as well.

lib/brain.js
lidio601: noted

lib/brain.js
lidio601: noted

lib/brain.js
lidio601: noted

lib/brain.js
```diff
@@ -605,7 +605,8 @@ Brain.prototype._connectedBrowser = function _connectedBrowser () {
     coins: this.trader.coins,
     twoWayMode: this.twoWayMode(),
     wifiList: wifiList,
-    rates
+    rates,
+    terms: this.trader.terms
```
joshmh: Will this update if the operator changes the terms screen on the backend?

lib/brain.js
```diff
@@ -760,6 +761,10 @@ Brain.prototype._processReal = function _processReal (req) {
       return this.lightningMode(true)
     case 'lightningNo':
       return this.lightningMode(false)
+    case 'termsAccepted':
+      this.fastUpdateTx({termsAccepted: true})
```
joshmh: The problem here is that the tx is not yet ready for sending to the server, as there are missing fields leading to a server error. Since we don't actually have to send the tx to the server yet, we should replace with something closer to your original change:

```
      this.tx = Tx.update(this.tx, {termsAccepted: true})
```

lib/brain.js
joshmh: Probably need to do this in the `sendRates()` function. Since it's becoming more general than just sending rates, it should be renamed.

lidio601: cool, ok

lib/brain.js
lidio601: got it, I would put those in a function thought

PR #116, opened 2018-06-08
mocked dispenser + dispense error handling

lidio601:
- mocked dispenser allows an extra configuration to force the balance
- improved brain.js behavior to signal dispense errors

PR #118, opened 2018-06-16
cash dispenser bugfix

lidio601:
I previously added an exception [here](https://github.com/lamassu/lamassu-machine/pull/116#discussion-diff-193978372R2572)

Actually I'm not sure if it's the proper way to handle this.
If the physical dispense is successful we transit from fiatComplete to goodbye screen but in the middle there is a delay of 1 minute, in which a user can start another transaction.

So in this case it make sense to check for:
```
const doComplete = this.state === 'fiatComplete' && this.tx.id === txId
```

To decide whenever to show or not the goodbye screen.

Anyway I think we should move the `emit('billDispenserCollected')` up before the delay so that we always run that, regardless the above check

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -2531,28 +2531,6 @@ Brain.prototype._dispense = function _dispense () {
     })
 }
 
-Brain.prototype.waitForCollection = function waitForCollection () {
-  const txId = this.tx.id
-
-  const checkCollection = () => {
-    const checkState = this.tx.id === txId && this.state === 'fiatComplete'
-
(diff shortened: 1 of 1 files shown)


lidio601:
[_dispense Error following successful dispense](https://trello.com/c/RQs2QrwX/536-dispense-error-following-successful-dispense)

lidio601:
https://github.com/lamassu/lamassu-machine/blob/33d25099aacd2c6f0f354b88ad363bc36ba2e694/lib/brain.js#L2584

PR #122, opened 2018-06-22
SMS compliance flow fix

lidio601:
Here are the fixes as suggested in https://trello.com/c/aNqpwqze/547-smsflowhandlereturnstate-logic-error

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -1382,15 +1382,20 @@ Brain.prototype.complianceAmount = function complianceAmount () {
     : this.minimumFiat()
 
   /**
+   * consider the last rejected bill
+   * to pick all the relevant tiers to run
+   */
+  const lastRejectedBill = tx.direction === 'cashIn'
+    ? _.defaultTo(BN(0), this.lastRejectedBillFiat)
+    : BN(0)
+
+  /**
    * Retrieve the user daily volume if any
    */
   const currentDailyVolume = this.getUsersDailyVolume()
 
   return BigNumber.max(currentDailyVolume, minimumTx)
-    /**
-     * add the inserted bill to the count
-     */
-    .add(this.lastRejectedBillFiat)
+    .add(lastRejectedBill)
 }
 
 /**
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -1292,39 +1292,31 @@ Brain.prototype.isTierCompliant = function isTierCompliant (tier) {
   }
 }
 
-Brain.prototype.smsFlowHandleReturnState = function smsFlowHandleReturnState () {
-  const amount = this.getUsersDailyVolume().add(this.lastRejectedBillFiat)
-  const nonCompliant = this.nonCompliantTiers(amount)
-
-  const otherTiers = _.isNil(this.failedCompliance) && !_.isEmpty(nonCompliant)
-
-  if (otherTiers) {
-    return complianceTiers.run(_.head(nonCompliant))
-  }
-
-  this.complianceReturnState()
-}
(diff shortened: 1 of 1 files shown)


lib/brain.js
```diff
@@ -1251,6 +1251,7 @@ Brain.prototype.isTierCompliant = function isTierCompliant (tier) {
     case 'sms':
       return !_.isNil(tx.phone)
     case 'hardLimit':
+      this.failedCompliance = 'hardLimit'
```
lidio601: In think this is not right because it may override other failed compliance

lib/brain.js
```diff
@@ -1267,10 +1267,12 @@ Brain.prototype.smsFlowHandleReturnState = function smsFlowHandleReturnState ()
...
   }
 
-  const amount = this.getUsersDailyVolume().add(this.lastRejectedBillFiat)
+  const amount = this.complianceDailyVolumeAmount()
   const nonCompliant = this.nonCompliantTiers(amount)
 
-  const otherTiers = _.isNil(this.failedCompliance) && !_.isEmpty(nonCompliant)
+  const otherTiers = _.isNil(this.failedCompliance) &&
+    !_.isEmpty(nonCompliant) &&
+    _.head(nonCompliant) !== 'hardLimit'
```
lidio601: I moved to this check instead to skip the `return complianceTiers.run(_.head(nonCompliant))` with _hardLimit_ that would crash

joshmh:
Thanks @lidio601, might take me a bit longer to review since I really need to remember how this code works.

lib/brain.js
joshmh: But if there's a hardLimit, won't it just skip compliance? Where will it stop the tx if the user has exceeded hardLimit?

lidio601: This currently throw [here](https://github.com/lamassu/lamassu-machine/blob/v3/lib/compliance-tiers.js#L40)

lib/brain.js
```diff
@@ -1283,17 +1285,21 @@ Brain.prototype.minimumFiat = function minimumFiat () {
...
+Brain.prototype.complianceDailyVolumeAmount = function complianceDailyVolumeAmount () {
   const tx = this.tx
   const minimumTx = tx.direction === 'cashIn'
     ? _.defaultTo(BN(0), tx.minimumTx)
     : this.minimumFiat()
 
-  const amount = this.getUsersDailyVolume()
+  return this.getUsersDailyVolume()
     .add(this.lastRejectedBillFiat)
     .add(minimumTx)
```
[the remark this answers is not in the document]
lidio601: sorry I can't see what do you mean. at 1295 we're adding minimumTx ok, but I can't see where we're considering the current tx balance.

lib/brain.js
joshmh: Right but that's never supposed to happen. Hitting the hardLimit is supposed to happen and shouldn't throw an error, it should make the user complete the transaction and not allow any more bills.

lidio601:
sorry it's not easy to see in "files changed" tab.

I moved complianceReturnState and smsFlowHandleReturnState since the first were only called by smsFlowHandleReturnState.
Now they use the same method to compute the _complianceAmount_.

I'm not sure whenever the complianceAmount is correct, btw now the consider the same limit.

joshmh:
@lidio601 Looks like this has conflicting changes. Can you rebase on top of latest dev?

lidio601:
done =)

joshmh:
Thanks! Added a couple more comments.

PR #125, opened 2018-07-08
minor fixes

lidio601:
While debugging https://trello.com/c/RQs2QrwX/536-dispense-error-following-successful-dispense I came up with those fixes that aren't directly related

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -522,17 +522,22 @@ Brain.prototype.scanPDF = function scanPDF () {
 }
 
 Brain.prototype.authorizeIdCardData = function () {
-  this.clearTimeoutToScanPDFCancel()
+  return Promise.resolve()
+    .then(() => {
+      this.clearTimeoutToScanPDFCancel()
 
-  const customer = this.customer
-  const data = idCardData.getData()
-  this.trader.updateCustomer(customer.id, {idCardData: data})
-    .then(result => {
-      const customer = result.customer
-      this.customer = customer
-      idCardData.dispatch('AUTHORIZED')
+      const customer = this.customer
+      const data = idCardData.getData()
+      return this.trader.updateCustomer(customer.id, {idCardData: data})
     })
-    .catch(() => {
+    .then(result => {
+      this.customer = result.customer
+      idCardData.dispatch('AUTHORIZED')
+    }, err => {
+      this._fiatError(err)
+    })
+    .catch(err => {
+      console.log('authorizeIdCardData error', err)
       idCardData.dispatch('BLOCKED_ID')
     })
 }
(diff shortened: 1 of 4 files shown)


lib/brain.js
```diff
@@ -522,15 +522,19 @@ Brain.prototype.scanPDF = function scanPDF () {
...
 
 Brain.prototype.authorizeIdCardData = function () {
-  this.clearTimeoutToScanPDFCancel()
+  Promise.resolve()
+    .then(() => {
+      this.clearTimeoutToScanPDFCancel()
 
-  const customer = this.customer
-  const data = idCardData.getData()
-  this.trader.updateCustomer(customer.id, {idCardData: data})
```
lidio601: this was raising an error outside the promise context because customer were undefined.
the result was that the UI got stuck

ui/src/app.js
```diff
@@ -505,7 +505,7 @@ $(document).ready(function () {
 })
 
 function targetButton (element) {
-  var classList = element.classList
+  var classList = element.classList || []
```
lidio601: I noticed this js error when switching the language via the language menu.

btw I noticed that the button on the top left of the screen is clickable even if there are not languages to switch (only english) and the button itself is hidden.

lib/brain.js
```diff
@@ -522,15 +522,19 @@ Brain.prototype.scanPDF = function scanPDF () {
...
+    })
     .then(result => {
-      const customer = result.customer
-      this.customer = customer
+      this.customer = result.customer
       idCardData.dispatch('AUTHORIZED')
+    }, err => {
+      this._fiatError(err)
     })
     .catch(() => {
```
joshmh: We should log the error here.

ui/src/app.js
joshmh: Cool, can you fix that, as well?

lib/brain.js
lidio601: it's already logged inside _fiatError

ui/src/app.js
lidio601: done.
I checked the functionality with 1, 2, 3 languages and it seems working fine to me.
I also fixed the css

ui/src/app.js
```diff
@@ -761,20 +762,9 @@ function setPrimaryLocales (primaryLocales) {
...
-    $('.js-two-language').show()
-  }
-
-  if (primaryLocales.length > 2) {
-    $('.js-multi-language').show()
-    $('.js-two-language').hide()
-  }
+  $('.js-menu-language').toggleClass('hide', sortedPrimaryLocales.length <= 1)
+  $('.js-multi-language').toggleClass('hide', sortedPrimaryLocales.length === 2)
+  $('.js-two-language').toggleClass('hide', sortedPrimaryLocales.length > 2)
```
joshmh: Nice :)

lidio601:
@joshmh ready for review

PR #129, opened 2018-07-26
lamp scan bay naming fix

lidio601:
I'm not sure if the correct name should be `hasNewScanBay` or `hasOldScanBay`


lidio601:
[Enable lamp on scan ID](https://trello.com/c/bTonY62u/526-enable-lamp-on-scan-id)

PR #137, opened 2018-08-24
Hardlimit error fix

lidio601:


lidio601:
['Unsupported tier: hardLimit at pick' - Hard limit alone invokes SMS and fails transaction](https://trello.com/c/SX1kf5kb/582-unsupported-tier-hardlimit-at-pick-hard-limit-alone-invokes-sms-and-fails-transaction)

PR #139, opened 2018-08-28
Update install instructions with preliminaries for MacOS and Ubuntu

kokipsiho:


naconner:
Thanks. :+1: 

By the way, there's also: https://github.com/lamassu/lamassu-machine/wiki/Running-v3-locally-on-Debian-or-Ubuntu

joshmh:
@naconner Yep, but I think we should keep it all in Install.md. It gets versioned with the software and there's just one place to look.

INSTALL.md
```diff
@@ -1,3 +1,35 @@
...
+## Packages
+
+```
+sudo apt-get update
+sudo apt-get install build-essential cmake libgtk2.0-dev pkg-config libavcodec-dev libavformat-dev libswscale-dev libpcsclite-dev
+```
+
+## OpenCV
+
+Download the OpenCV source archive and unpack it. In source folder run following:
```
joshmh: Let's add a link to the source archive or at least to the releases page.

kokipsiho: Added link to releases page.

PR #141, opened 2018-09-01
smsVerification state added

lidio601:


lidio601:
[No brain state exists for ID verification screen](https://trello.com/c/FwTAeNTY/591-no-brain-state-exists-for-id-verification-screen)

lib/brain.js
```diff
@@ -1951,6 +1951,9 @@ Brain.prototype._billRead = function _billRead (data) {
   if (!this.isCompliant(amount)) {
     // Cancel current bill
     this.billValidator.reject()
+
+    this._setState('smsVerification')
+
```
joshmh: Should use ``_transitionState`` instead of ``_setState`` and ``browser().send``.

PR #146, opened 2018-09-19
Stop redeem from getting "Out of cash" errors

RafaelTaranto:
Redeem was checking available cash, which was causing a
"Out of cash" error even though the cash is already reserved on l-s.

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -2409,7 +2409,9 @@ Brain.prototype._chooseFiat = function _chooseFiat () {
   this.tx = Tx.update(this.tx, {fiatCode: this.fiatCode, direction: 'cashOut'})
 
   const response = this._getFiatButtonResponse()
-  if (response.activeDenominations.isEmpty) return this._timedState('outOfCash')
+
+  // No need to check available cash on redeem. It is already reserved on l-s.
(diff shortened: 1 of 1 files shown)


RafaelTaranto:
[Trello card](https://trello.com/c/5gqWVkG6/607-unable-to-dispense-if-low-on-cash)

lib/brain.js
```diff
@@ -2409,7 +2409,9 @@ Brain.prototype._chooseFiat = function _chooseFiat () {
   this.tx = Tx.update(this.tx, {fiatCode: this.fiatCode, direction: 'cashOut'})
 
   const response = this._getFiatButtonResponse()
-  if (response.activeDenominations.isEmpty) return this._timedState('outOfCash')
+
+  // No need to check available cash on redeem. It is already reserved on l-s.
+  if (response.activeDenominations.isEmpty && !this.redeem) return this._timedState('outOfCash')
```
joshmh: Hmm, this is simple to implement but shouldn't we be fixing the original problem that we were double counting instead of giving up on the check entirely? I'm not sure we should completely rely on the l-s reservation.

RafaelTaranto:
Looks like this is a unrelated error and the PR not be applied, more on the trello card.

PR #148, opened 2018-09-20
Stop machine from hanging when verifying ID

RafaelTaranto:
Declining to verify ID during a cash_in was causing the machine to
hang indefinedely.


RafaelTaranto:
[Trello Card](https://trello.com/c/Fkvs0KQT/605-tx-fails-machine-hangs-after-declining-to-verify-amount-overcredited-after-verifying-id)

PR #149, opened 2018-09-20
Stop overcrediting after ID verification

RafaelTaranto:
The machine was counting a bill that gets returned to the user when
asking for ID verification on cash_in


RafaelTaranto:
[Trello Card](https://trello.com/c/Fkvs0KQT/605-tx-fails-machine-hangs-after-declining-to-verify-amount-overcredited-after-verifying-id)

PR #151, opened 2018-09-22
id-card-upload stm flow fix

lidio601:
I spotted this flow error when coming from a successful id card upload tier


lidio601:
[Upload ID image to l-s](https://trello.com/c/aiRzeg17/288-upload-id-image-to-l-s)

PR #153, opened 2018-09-24
Add "Are you sure?" screen

RafaelTaranto:
Some user actions can impair redepmtion or cancel transactions, this
screen was added to get a confirmation from the user.

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -43,6 +43,7 @@ const NON_TX_STATES = ['networkDown', 'connecting', 'wifiConnected', 'pairing',
   'initializing', 'booting']
 const ARE_YOU_SURE_ACTIONS = ['cancelTransaction', 'continueTransaction']
 const ARE_YOU_SURE_HANDLED = ['depositCancel']
+const ARE_YOU_SURE_SMS_HANDLED = ['cancelPhoneNumber', 'cancelSecurityCode']
 const ARE_YOU_SURE_HANDLED_SMS_COMPLIANCE = ['deposit_timeout', 'rejected_zero_conf']
 const INITIAL_STATE = 'start'
 const MIN_SCREEN_TIME = 1000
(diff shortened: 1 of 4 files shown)


RafaelTaranto:
[Add 'Are you sure?' screen](https://trello.com/c/HHcawvND/562-add-are-you-sure-screen)

ui/start.html
```diff
@@ -1327,6 +1327,22 @@ <h1 class="js-i18n js-terms-title"></h1>
         </section>
       </section>
 
+      <section class="viewport are_you_sure_state" data-tr-section="are-you-sure">
+        <section class="info">
+          <h1 class="js-i18n">Are You Sure?</h1>
```
RafaelTaranto: I'm not sure how i18n is actually working on this project. Should I add the texts on this new screen to some file?

lib/brain.js
```diff
@@ -2727,6 +2749,35 @@ Brain.prototype._fiatReceipt = function _fiatReceipt () {
   })
 }
 
+Brain.prototype.areYouSureHandled = function areYouSureHandled (action) {
+  return _.includes(action, ARE_YOU_SURE_HANDLED) || 
+    (action === 'cancelPhoneNumber' && _.includes(this.complianceReason, ARE_YOU_SURE_HANDLED_SMS_COMPLIANCE))
+}
+
+Brain.prototype.areYouSure = function areYouSure () {
+  this._timedState('areYouSure')
```
RafaelTaranto: Should this have a timeout other than the default one?

lib/brain.js
```diff
@@ -1262,7 +1281,10 @@ Brain.prototype.smsCompliance = function smsCompliance (opts = {}) {
     this.flow = null
     this.failedCompliance = 'sms'
 
-    if (returnState) return this.smsFlowHandleReturnState()
+    if (this.returnState && !_.includes(this.complianceReason, ARE_YOU_SURE_HANDLED_SMS_COMPLIANCE)) {
```
RafaelTaranto: I'm not sure if using the local returnReturn state was coding style or if it is to stop another method to change the global one, I could add another global variable instead of messing with this.

ui/start.html
joshmh: No need to add texts to a file. The ``js-i18n`` class marks this as a screen to be translated.

lib/brain.js
joshmh: Default one is fine.

PR #155, opened 2018-09-25
Fix unresponsive start screen on screen timeout

RafaelTaranto:
Start screen got unresponsive after a timeout during a smsCompliance
flow, the timeout was not cleaning up the flow handler which in turn
swallowed all requests.


RafaelTaranto:
[Unresponsive start screen](https://trello.com/c/6FfSgtMH/625-unresponsive-start-screen)

PR #160, opened 2018-10-09
cleanup error handling in camera wrapper

lidio601:
I cleanup up how I was handling the error.
It was pretty redundant.

I'm still now sure about this approach:
Basically you come with a single event from the camera which is splitted in 3:
- onError
- onFaceDetected
- onFrame
and you can pass the handler you're interested in.

Although we're not really catching exceptions raised from the native part, I needed to introduce the error handling because of the mocked version.


lidio601:
[id-card-photo: 'No such file or directory' on camera initialisation](https://trello.com/c/qU2aNL0J/650-id-card-photo-no-such-file-or-directory-on-camera-initialisation)

lidio601:
[id-card-photo: Scanning ID immediately exits due to no face](https://trello.com/c/YAgsfpXC/654-id-card-photo-scanning-id-immediately-exits-due-to-no-face)

lidio601:
Further fixes: now I'm getting the camera settings from device_config.json like it is for the QR Code scanner and the barcode/idCard scanner.
I'm using idCard settings as fallback in case photoCard settings is missing within device_config.json.

@naconner you can also use bin/cam.js to test the camera-wrapper functionality/settings.

lidio601:
@joshmh I tried to align the _lib/mocked/scanner.js_ and _lib/scanner.js_ code.
I still don't like the fact that we're kind of duplicating the code here, but I didn't found a better way for now.
It would require more refactoring around that but it's not really worth it for now.

Also if we're going to move the barcode scanner to use the native camera-wrapper we can get rid of the mocked scanner.

The camera-wrapper uses internally a mocked version of the native module if you specify:

```
{
  "brain": {
    ...
    "mockCameraImage": "mock_data/compliance/license.jpg"
    ...
  },
```

in your _device_config.json_. In this way we can test the whole js code without depending on the platform.

bin/cam.js
```diff
@@ -1,14 +1,15 @@
-const cam = require('../lib/camera')
+// const scanner = require('../lib/scanner')
+const scanner = require('../lib/mocks/scanner')
```
joshmh: Why is using the mock scanner here?

bin/cam.js
```diff
@@ -1,14 +1,15 @@
-const cam = require('../lib/camera')
+// const scanner = require('../lib/scanner')
```
joshmh: Instead of commenting, just delete.

lib/camera/examples/example.js
```diff
@@ -4,17 +4,39 @@ var camera = require('../')
...
+  singleShot: false,
+
+  // width: 0,
+  // height: 0,
+
+  // width: undefined,
+  // height: undefined,
+
+  // width: 640,
+  // height: 480,
```
joshmh: Explain what the different values do.

lib/camera/index.js
```diff
@@ -10,8 +10,8 @@ var defaultOpts = {
   // input: '0',
   // codec: '.jpg',
   // frameCallback: function (err, frameRaw, faceDetected) {},
-  // onFrame: function (err, frameRaw) {},
-  // onFaceDetected: function (err, frameRaw) {},
+  // onFrame: function (frameRaw) {},
+  // onFaceDetected: function (frameRaw) {},
```
joshmh: Why are all these defaults commented out?

lib/mocks/scanner.js
```diff
@@ -1,73 +1,88 @@
...
-  scanMainQR: scanMainQR,
-  scanPDF417: scanPDF417,
-  scanPhotoCard: scanPhotoCard,
-  cancel: cancel
+  config,
+  scanPairingCode,
+  scanMainQR,
+  scanPDF417,
+  scanPhotoCard,
+  cancel
```
joshmh: 👍

lib/scanner.js
```diff
@@ -1,24 +1,23 @@
...
-var cancelFlag = false
-
+const v4l2camera = require('v4l2camera')
+const jpg = require('jpeg-turbo')
+const manatee = require('manatee')
+const Pdf417Parser = require('./compliance/parsepdf417')
+const coinUtils = require('./coins/utils')
+
+let configuration = null
+let cancelFlag = false
```
joshmh: 👍

bin/cam.js
lidio601: I just need to call the `scanPhotoCard` method which is the same between mocked and un-mocked version.
Unfortunately I cannot use the un-mocked version on my dev environment as I'm missing v4l2camera and it would raise an exception

bin/cam.js
lidio601: ok

lib/camera/index.js
lidio601: to serve as documentation.
I added also the JSdoc above so in theory I can remove those

lib/camera/examples/example.js
lidio601: added the JSdoc above

lidio601:
closed in favour of #164

PR #162, opened 2018-10-09
Add new F56 currency support

naconner:
Adds MYR. Leaving open and will add additional currencies soon.


naconner:
[Add MYR, HKD, ISL to F56 bills](https://trello.com/c/NcIVsBXu/672-add-myr-hkd-isl-to-f56-bills)

PR #166, opened 2018-10-15
Handle no SOH when in waitEOT state

kokipsiho:


joshmh:
[No 'start of header' response from Puloon leaves machine in persistent 'booting' state](https://trello.com/c/1dfZXSRg/576-no-start-of-header-response-from-puloon-leaves-machine-in-persistent-booting-state)

PR #167, opened 2018-10-17
Add support for bech32 BTC address, unit test validators

kokipsiho:
Add new line to unit tests

### DIFF
--- a/lib/coins/btc.js
+++ b/lib/coins/btc.js
@@ -1,6 +1,20 @@
-const bs58check = require('bs58check')
-const bech32 = require('bech32')
 const _ = require('lodash/fp')
+const base58Validator = require('./validators').base58Validator
+const bech32Validator = require('./validators').bech32Validator
+
+const base58Opts = {
+  bufferLength: 21,
+  mainNetPrefix: [0x00, 0x05],
+  testNetPrefix: [0x6f, 0xc4]
+}
+
+const bech32Opts = {
+  witnessVersion: 0,
+  minAddressLen: 20,
+  maxAddressLen: 32,
+  mainNetPrefix: 'bc',
+  testNetPrefix: 'tb'
+}
 
 module.exports = {depositUrl, parseUrl, formatAddress}
 
@@ -35,54 +49,10 @@ function formatAddress (address) {
   return address
 }
 
-function base58Validator (network, address) {
-  try {
-    const buf = bs58check.decode(address)
-    const addressType = buf[0]
-
-    if (buf.length !== 21) {
-      console.log(`Invalid base58 address length: ${buf.length}`)
-      return false
-    }
-
-    if (network === 'main' && _.includes(addressType, [0x00, 0x05])) return true
-    if (network === 'test' && _.includes(addressType, [0x6f, 0xc4])) return true
-
-  } catch (error) {
-    console.log('Failed to decode base58 Bitcoin address')
-    return false
-  }
-}
-
-function bech32Validator (network, address) {
-  let decoded
-  try {
-    decoded = bech32.decode(address)
-  } catch (error) {
-    console.log('Failed to decode bech32 Bitcoin address')
-    return false
-  }
-
-  const witnessVersion = decoded.words[0]
-  if (witnessVersion < 0 || witnessVersion > 16) {
-    console.log('Wrong witness version for bech32')
-    return false
-  }
-
-  const data = bech32.fromWords(decoded.words.slice(1))
-  if (data.length !== 20 && data.length !== 32) {
-    console.log(`Invalid bech32 address length: ${data.length}`)
-    return false
-  }
-
-  if (network === 'main' && decoded.prefix === 'bc') return true
-  if (network === 'test' && decoded.prefix === 'tb') return true
-}
-
 function validate (network, address) {
   if (!network) throw new Error('No network supplied.')
   if (!address) throw new Error('No address supplied.')
-  if (base58Validator(network, address)) return true
-  if (bech32Validator(network, address)) return true
+  if (base58Validator(network, address, base58Opts)) return true
+  if (bech32Validator(network, address, bech32Opts)) return true
   return false
 }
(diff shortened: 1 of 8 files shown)


lib/coins/btc.js
```diff
@@ -34,23 +35,53 @@ function formatAddress (address) {
...
-    if (!network) throw new Error('No network supplied.')
-    if (!address) throw new Error('No address supplied.')
-
     const buf = bs58check.decode(address)
     const addressType = buf[0]
 
-    if (buf.length !== 21) throw new Error(`Invalid length: ${buf.length}`)
+    if (buf.length !== 21) {
+      console.log(`Invalid base58 address length: ${buf.length}`)
+    }
```
joshmh: Shouldn't we return ``false`` here?

```suggestion
      return false      
    }
```

lib/coins/zec.js
```diff
@@ -21,22 +30,26 @@ function depositUrl (address, amount) {
...
+
+  //   if (network === 'main' && _.includes(addressType, MAIN_PREFIXES)) return true
+  //   if (network === 'test' && _.includes(addressType, TEST_PREFIXES)) return true
+
+  //   throw new Error('General error')
+  // } catch (err) {
+  //   console.log(err)
+  //   console.log('Invalid address: [%s] %s', network, address)
+  //   return false
+  // }
```
joshmh: Remove comment.

lib/coins/zec.js
```diff
@@ -21,22 +30,26 @@ function depositUrl (address, amount) {
...
-    throw new Error('General error')
-  } catch (err) {
-    console.log(err)
-    console.log('Invalid address: [%s] %s', network, address)
-    return false
-  }
+  if (!network) throw new Error('No network supplied.')
+  if (!address) throw new Error('No address supplied.')
+  if (base58Validator(network, address, base58Opts)) return true
+  return false
```
joshmh:
```javascript
  return base58Validator(network, address, base58Opts)
```

test/unit/test-crypto-validation.js
```diff
@@ -0,0 +1,92 @@
+import test from 'ava'
+import crypt0Validator from '../../lib/coins/validators'
```
joshmh: Should be ``cryptoValidator``.

test/unit/test-crypto-validation.js
```diff
@@ -0,0 +1,92 @@
...
+
+test('Should validate BTC bech32', t => {
+  t.plan(2)
+  const mainNetaddr = 'bc1qar0srrr7xfkvy5l643lydnw9re59gtzzwf5mdq'
+  const testNetaddr = 'tb1qw508d6qejxtdg4y5r3zarvary0c5xw7kxpjzsx'
+  const validatedMain = crypt0Validator.bech32Validator('main', mainNetaddr, BTCBech32Opts)
+  const validatedTest = crypt0Validator.bech32Validator('test', testNetaddr, BTCBech32Opts)
+  t.true(validatedMain)
+  t.true(validatedTest)
+})
```
joshmh: Very cool! But should also test failure of invalid addresses of different kinds, at least all the things we're checking in the code, like address length.

lib/coins/validators/index.js
```diff
@@ -1,61 +1,35 @@
...
     if (buf.length !== opts.bufferLength) {
       console.log(`Invalid base58 address length: ${buf.length}`)
       return false
     }
 
-    if (network === 'main' && validatePrefix(opts.mainNetPrefix)) return true
-    if (network === 'test' && validatePrefix(opts.testNetPrefix)) return true
+    if (network === 'main' && validatePrefix(opts.mainNetPrefix, buf)) return true
+    if (network === 'test' && validatePrefix(opts.testNetPrefix, buf)) return true
     throw new Error('General error')
```
joshmh: Should log that network isn't recognized and return false.

lib/coins/validators/index.js
```diff
@@ -1,61 +1,35 @@
...
 
   } catch (error) {
     console.log('Failed to decode base58 address')
```
joshmh: Also print out the caught error message for more debug info.

test/unit/test-crypto-validation.js
```diff
@@ -59,34 +59,51 @@ test('Should validate DASH P2SH', t => {
...
+
+test('Should fail length validation', t => {
+  const addr = '1BvBMSEYstWetqTFn5Au4m4GFg7xJaNVN2'
+  const opts = BTCBase58Opts
+  opts.bufferLength = 10
+  const validated = cryptoValidator.base58Validator('main', addr, opts)
+  t.false(validated)
+})
+
+test('Should throw exception on invalid address', t => {
```
joshmh: Document what's invalid about the address.

lib/coins/btc.js
```diff
@@ -34,23 +41,30 @@ function formatAddress (address) {
...
-    throw new Error('General error')
-  } catch (err) {
-    console.log(err)
-    console.log('Invalid bitcoin address: [%s] %s', network, address)
+  const witnessVersion = decoded.words[0]
+  if (witnessVersion !== 0) {
+    console.log('Unsupported witness version for bech32')
     return false
   }
+
```
joshmh: @kokipsiho I think you misunderstood my comment. I just wanted the length checking code to be outside the if/else block, I didn't want to get rid of it completely. You probably wondered why I would want to get rid of the check. If so, please ask about things like this, or we risk losing important checks. I'm sure there will be a lot of things I don't catch in the future, so we both have to be looking at this carefully.
 
```javascript
  const data = bech32.fromWords(decoded.words.slice(1))	
  if (data.length !== 20 && data.length !== 32) {	
    console.log(`Invalid bech32 address length: ${data.length}`)	
    return false	
  }
```

kokipsiho: I think what confused me is that length is determined by witness version. In future I will ask when in doubt.

joshmh: 👍

PR #168, opened 2018-10-18
don't repeat sms tier if already verified

lidio601:


lidio601:
I think rafael already fixed this, not sure were the fix is gone

lidio601:
[id-card-photo tier on cash-out requires second entry of SMS](https://trello.com/c/HEVLboEJ/651-id-card-photo-tier-on-cash-out-requires-second-entry-of-sms)

PR #170, opened 2018-10-22
Actionemitter fix

kokipsiho:


lib/ssuboard/process-door.js
```diff
@@ -1,9 +1,8 @@
 const actionEmitter = require('../action-emitter')
-const util = require('util')
 
 let isDoorSecured = null
 
-actionEmitter.isDoorSecured = () => isDoorSecured
+processDoor.isDoorSecured = () => isDoorSecured
```
joshmh: I don't think this addresses the problem. There are a few places in other files that reference ``actionEmitter.isDoorSecured``. We want to fix those references so that they reference processDoor or some other relevant module and not actionEmitter.

PR #174, opened 2018-10-23
Fix 'are you sure' screen idle process

RafaelTaranto:
When going idle the are you sure screen has to perform a couple of steps
before calling the '_idle' function.


joshmh:
Thanks for fixing formatting and splitting into two commits!

PR #176, opened 2018-10-24
Change 'Are You Sure' buttons to the standard

RafaelTaranto:


RafaelTaranto:
[areYouSure screen's buttons unresponsive](https://trello.com/c/tkpNUCoc/684-areyousure-screens-buttons-very-slow-to-respond)

ui/css/main.css
```diff
@@ -2000,6 +2000,10 @@ body  { margin: 0;}
   text-align: center !important;
 }
 
+.info .button.are-you-sure-button {
```
joshmh: This isn't critical but in the future we should keep css selectors flat (i.e., just ``.are-you-sure-button``) because deep selectors impact performance.

RafaelTaranto: I started doing just `.are-you-sure-button` but it was loosing in priority to `.info .button`.

PR #178, opened 2018-10-25
Add CSS for 1920x1080 resolution

RafaelTaranto:


ui/start.html
```diff
@@ -4,7 +4,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="stylesheet" href="flag-icon-css/css/flag-icon.css">
     <link rel="stylesheet" href="css/fonts.css">
-    <link rel="stylesheet" href="css/main.css">
+    <link rel="stylesheet" href="css/main.css" media="screen and (max-height: 800px)">
```
RafaelTaranto: This could be annoying to testers and developers, I'm open to suggestions on a better way to do it! I'll do some tests setting the resolution change to 1080 instead of 800, but I was having some issues earlier.

RafaelTaranto:
[CSS for 1920x1080 resolution.](https://trello.com/c/xuHqlXGm/667-css-for-1920x1080-resolution)

ui/start.html
```diff
@@ -4,7 +4,8 @@
     <meta name="viewport" content="width=device-width, initial-scale=1.0">
     <link rel="stylesheet" href="flag-icon-css/css/flag-icon.css">
     <link rel="stylesheet" href="css/fonts.css">
-    <link rel="stylesheet" href="css/main.css">
+    <link rel="stylesheet" href="css/main.css" media="screen and (max-height: 1080px)"> 
+    <link rel="stylesheet" href="css/main-1080.css" media="screen and (min-height: 1080px)">
```
joshmh: Which rule is chose if the height is exactly 1080?

RafaelTaranto: Oh, it would load them both, but the 1080 one would have a higher priority. Should I reduce some pixels off of max-height?

PR #182, opened 2018-11-02
fix pdf417 for colorado

RafaelTaranto:
Milestone: 7.3.x Crafty Chnemu


RafaelTaranto:
[Colorado ID parsing not carried over from legacy to next-gen](https://trello.com/c/DXBLbRaD/699-colorado-id-parsing-not-carried-over-from-legacy-to-next-gen)

PR #189, opened 2018-11-21
Show batch counter ui only for multibatch dispense

kokipsiho:


ui/src/app.js
```diff
@@ -1206,10 +1206,18 @@ function fiatComplete (tx) {
...
   setState('fiat_complete')
 }
 
+function toggleBatchCounter(state) {
+  const opacity = state === 'show' ? 100 : 0
+  $('.batch').css('opacity', opacity)
+}
+
 function dispenseBatch (data) {
+  if (data.of > 1) toggleBatchCounter('show')
```
joshmh: It's safer to hide here for batches of 1, otherwise we're assuming that fiatComplete has already run since the last toggle.

kokipsiho: Makes sense

ui/src/app.js
```diff
@@ -1206,10 +1206,18 @@ function fiatComplete (tx) {
...
   setState('fiat_complete')
 }
 
+function toggleBatchCounter(state) {
+  const opacity = state === 'show' ? 100 : 0
+  $('.batch').css('opacity', opacity)
+}
+
 function dispenseBatch (data) {
+  if (data.of === 1) toggleBatchCounter('hide')
```
joshmh: I meant that you should handle both cases: hide if ``data.of === 1`` and show if ``data.of > 1``.

test/unit/test-optimize-batch.js
```diff
@@ -35,3 +36,10 @@ test('Sum of optimized batches equals input', t => {
 
   t.is(_.sum(_.flatten(o)), 66)
 })
+
+test('Should return input array', t => {
+  const input = [7,15]
+  const o = optimizeDispenseWithoutLimit(input)
+  
+  t.deepEqual(o, input)
+})
```
RafaelTaranto: This test will fail after https://github.com/lamassu/lamassu-machine/pull/195 gets merged. It should return [[7,15]].

RafaelTaranto:
@joshmh this PR can be closed without merging!

PR #195, opened 2018-12-05
Fix dispense optimizer no limit

RafaelTaranto:


RafaelTaranto:
[PuloonDispenseError: Cannot set property 'err' of undefined](https://trello.com/c/R6Vq7ndl/715-puloondispenseerror-cannot-set-property-err-of-undefined)

PR #196, opened 2018-12-07
Pagination on terms and conditions screen

DatoBula:


DatoBula:
[Paginate Terms & Conditions screen](https://trello.com/c/Rjg2HEbX/689-paginate-terms-conditions-screen)

ui/src/app.js
```diff
@@ -650,14 +656,135 @@ function setWifiList (recs, requestedPage) {
  * @param {String} data.cancel
  */
 function setTermsScreen (data) {
-  const $screen = $('.js-terms-screen')
-
+  var $screen = $('.js-terms-screen')
```
joshmh: This should remain ``const``.

ui/src/app.js
```diff
@@ -650,14 +656,135 @@ function setWifiList (recs, requestedPage) {
...
+      lineEndIndex = calculateLinesLastWordIndex(textArray, lineEndIndex + 1)
+    } else {
+      lineStartIndex = 0
+      lineEndIndex = calculateLinesLastWordIndex(textArray, lineEndIndex)
+    }
+    linesArray.push(createLine(textArray, lineStartIndex, lineEndIndex))
+  }
+  return linesArray
+}
+
```
joshmh: This splitting is problematic since we don't know how the browser will format the text. Instead, I think we should use the ``scrollTo()`` JavaScript method to just scroll the text down and up one screen (minus a line or two). The div with the text can have a fixed height, and have ``overflow: hidden``.

ui/src/app.js
```diff
@@ -650,14 +656,135 @@ function setWifiList (recs, requestedPage) {
...
+    makeButtonDisabled(buttonUp)
+  } else {
+    makeButtonEnabled(buttonUp)
+  }
+  if (currentPageIndex === border) {
+    makeButtonDisabled(buttonDown)
+  } else {
+    makeButtonEnabled(buttonDown)
+  }
+}
```
joshmh: Buttons should not appear at all (not even disabled) if text completely fits into the screen.

PR #201, opened 2018-12-09
Fix updater for ssuboard v1.2

littlenaut:


littlenaut:
[Set up updater for ssuboard](https://trello.com/c/xhQRPEnS/759-set-up-updater-for-ssuboard)

lib/update/updater.js
```diff
@@ -28,8 +33,16 @@ Updater.factory = function factory (config) {
...
+  const netInterfaces = os.networkInterfaces()
+  const firstWlanName = Object.keys(netInterfaces).filter(name => name.startsWith('wlan')).sort()[0]
+  const firstEthName = Object.keys(netInterfaces).filter(name => name.startsWith('eth')).sort()[0]
+  const mainInterfaceName = firstWlanName ? firstWlanName : firstEthName
+
+  if (mainInterfaceName) {
+    const macAddr = netInterfaces[mainInterfaceName][0].mac
+    return macAddr.replace(/:/g, '-')
+  }
+  else return null
```
joshmh: Consider just filtering ``netInterfaces`` for all ``mac`` that's not ``00:00:00:00:00:00``, sorting, and choosing the first one.

littlenaut: Thought a little bit about your suggestion. Sounds better than the current one. I also thought that we don't need to sort the result, but we do since there might not be any guarantees that the order in which they're reported is maintained across reboots.
Going for it!

lib/update/updater.js
```diff
@@ -19,7 +20,11 @@ var Updater = function (config) {
   this.extractor = require('./extractor').factory(this.config.extractor)
   this.downloading = false
   this.ack = null
+
   this.deviceId = this._fetchDeviceId()
+  if (this.deviceId == null) {
+    console.log('[WARN]: Device ID could not be retrived.')
```
joshmh: Fix spelling (retrived).

lib/update/updater.js
```diff
@@ -28,8 +33,30 @@ Updater.factory = function factory (config) {
...
+          macAddrs.push({ name: [key], mac: item.mac })
+      }
+    })
+  }
+
+  macAddrs.sort((lhs, rhs) => {
+      if (lhs.mac < rhs.mac) return -1
+      else if (lhs.mac > rhs.mac) return 1
+      else return 0
+  })
```
joshmh: Combine filter and sort from lodash/fp to handle this.

lib/update/updater.js
```diff
@@ -19,7 +20,11 @@ var Updater = function (config) {
   this.extractor = require('./extractor').factory(this.config.extractor)
   this.downloading = false
   this.ack = null
+
   this.deviceId = this._fetchDeviceId()
+  if (this.deviceId == null) {
```
joshmh: ``if (!this.deviceId)`` is probably safer, or consider throw and catching an exception.

lib/update/updater.js
```diff
@@ -35,24 +35,16 @@ Updater.factory = function factory (config) {
...
-
-  if (macAddrs.length > 0) {
-    const name = macAddrs[0].name
-    const mac = macAddrs[0].mac.replace(/:/g, '-')
+  const mainInterface =
+    _(netInterfaces)
+    .pickBy((item) => item[0].mac !== '00:00:00:00:00:00')
+    .toPairs()
+    .sortBy((item) => item[1].mac)
+    .first()
```
joshmh: After ``toPairs``, we get an array of pairs. The second half of each pair is an array, possibly with only one value. So I don't think the sort will work here.

Here's what I came up with, although it loses the interface name, which we don't really need:

```
_(os.networkInterfaces())
  .values()
  .flatten()
  .reject(_.isMatch({mac: '00:00:00:00:00:00'}))
  .sortBy('mac')
  .map('mac')
  .first()
```

PR #210, opened 2018-12-18
Send cashbox remove and cashbox insert events to lamassu-server

dabulashvili:


dabulashvili:
[Log event for validator cash box removal and reinserts](https://trello.com/c/Uplwgvlk/665-log-event-for-validator-cash-box-removal-and-reinserts)

PR #212, opened 2018-12-21
Remove word wrap from Verify ID photo screen

RafaelTaranto:


RafaelTaranto:
[Change word-wrap on scan_photo screen](https://trello.com/c/8mEdbfqg/789-change-word-wrap-on-scanphoto-screen)

PR #213, opened 2018-12-26
Increase thresholds to detect ID face

naconner:


naconner:
[Increase threshold for facial detection](https://trello.com/c/CUcXbngX/793-increase-threshold-for-facial-detection)

PR #217, opened 2018-12-30
Rename and improve l-m reset script

littlenaut:
- Renamed from os-cleanup.sh to reset.sh because we're only reseting
  state related with lamassu-machine.
- Bash's history cleanup only works as expected when executed from the
  last login shell.


littlenaut:
[Create ssuboard image from Soneco's base image](https://trello.com/c/njHJgIXt/796-create-ssuboard-image-from-sonecos-base-image)

PR #222, opened 2019-01-09
Remove retry command from dispense on puloon

RafaelTaranto:


RafaelTaranto:
[Remove retry command from puloon code](https://trello.com/c/W7sfRJ2E/804-remove-retry-command-from-puloon-code)

joshmh:
I'm thinking maybe we should just completely remove retry. With retry in here, we could send a new command while we're still retrying the old command which will confuse the Puloon even more. @RafaelTaranto what do you think?

RafaelTaranto:
That was my first idea but after going through the cards of the original
commit looked like the retry did fix something.

But it does feel like we should favor a more stable puloon code, yeah.

joshmh:
Ok, I'll merge for now and we can revisit later.

PR #223, opened 2019-01-11
Fix 800p/1080p media query for Surf 1.8

littlenaut:


littlenaut:
[Fix media query for Surf 1.8 in lamassu-machine](https://trello.com/c/i88sUfCI/807-fix-media-query-for-surf-18-in-lamassu-machine)

PR #231, opened 2019-01-19
Camera mock should not load if unneeded

RafaelTaranto:


RafaelTaranto:
[l-m camera-wrapper requires dev dependency](https://trello.com/c/ctOidZVB/831-l-m-camera-wrapper-requires-dev-dependency)

PR #238, opened 2019-01-30
 Add AUD & PHP, correct recent currency lengths for F56

naconner:


naconner:
[Additional F56 currencies](https://trello.com/c/1xCi1lfF/850-additional-f56-currencies)

PR #242, opened 2019-02-06
set updater.js keepAlive

naconner:
Allows deployment of `bb-debs` package following push of `codebase-7.2.1`'s update.tar and `updater.js` restart.

Via @littlenaut.


naconner:
[Allow debs package to deploy remotely](https://trello.com/c/d5hCI3bx/859-allow-debs-package-to-deploy-remotely)

PR #243, opened 2019-02-08
New register phone screen

RafaelTaranto:


RafaelTaranto:
[New machine-ui 'Phone Number' screen](https://trello.com/c/Uopd0rKz/848-new-machine-ui-phone-number-screen)

PR #251, opened 2019-02-28
Use commission rate from machine istead of server

RafaelTaranto:


RafaelTaranto:
[Log ticker rate at time of transaction](https://trello.com/c/6IpkHUXu/733-log-ticker-rate-at-time-of-transaction)

PR #254, opened 2019-03-05
Compliance: Paper wallet printing

dangerware:


littlenaut:
[Private key printing](https://trello.com/c/A9y5mQE1/808-private-key-printing)

littlenaut:
[Private key printing](https://trello.com/c/A9y5mQE1/808-private-key-printing)

lib/brain.js
```diff
@@ -1757,7 +1758,52 @@ Brain.prototype._start = function _start () {
...
+
+      this.trader.createCashInWallet(this.tx, (res, err) => {
+
+        if (err) {
+          // TODO: New screen to surface this error to the UI
+          return
+        }
+
+        const pubAddr = res.body.toAddress
+        const privAddr = res.body.toAddress
```
joshmh: Aren't these two the same?

lib/brain.js
```diff
@@ -1757,7 +1758,52 @@ Brain.prototype._start = function _start () {
...
+            // TODO: New screen to surface this error to the UI
+            return
+          }
+          this._handleScan(pubAddr)
+        })
+      })
+    })
+
+  }
+  else if (deviceConfig.complianceFeatures.printCashInWallet) {
```
joshmh: Shouldn't need ``else``, just return after previous ``if`` and start a new ``if``.

lib/brain.js
```diff
@@ -1757,7 +1758,52 @@ Brain.prototype._start = function _start () {
...
+    /* TODO: The machine has been incorrectly configured.
+       It is set to comply with wallet printing but it isn't
+       configured to use a kiosk printer to print the wallet.
+       Present an error screen to the user in order for it to
+       contact the Operator? */
+  }
+  else this._startAddressScan()
```
joshmh: Don't need ``else``, put ``return`` in previous ``if``.

lib/brain.js
```diff
@@ -1757,7 +1758,52 @@ Brain.prototype._start = function _start () {
...
   }
 
-  this._startAddressScan()
+  if (deviceConfig.complianceFeatures.printCashInWallet &&
+      deviceConfig.kioskPrinter &&
+      this.tx.cryptoCode === 'BTC') {
+
+    this._transitionState('cashInWaiting')
+
+    kioskPrinter.checkStatus(deviceConfig.kioskPrinter, (stats, err) => {
```
joshmh: Use promises.

lib/brain.js
```diff
@@ -1757,7 +1758,52 @@ Brain.prototype._start = function _start () {
...
+        if (err) {
+          // TODO: New screen to surface this error to the UI
+          return
+        }
+
+        const pubAddr = res.body.toAddress
+        const privAddr = res.body.toAddress
+        const printerCfg = deviceConfig.kioskPrinter
+
+        kioskPrinter.printWallet(pubAddr, privAddr, printerCfg, (err) => {
```
joshmh: Use promises.

lib/brain.js
```diff
@@ -1757,10 +1759,54 @@ Brain.prototype._start = function _start () {
...
+      })
+
+      return
+  }
+
   this._startAddressScan()
+
+  /* TODO: It's not exactly clear what the structure of this.tx will be when
+     it arrives at the browser. It will depend on the order that NodeJS decides
+     to execute the tasks. We should handle this deterministically. */
```
joshmh: Not clear what this means. Turn this into a Trello card.

lib/brain.js
```diff
@@ -1757,10 +1759,54 @@ Brain.prototype._start = function _start () {
...
+            this.tx = Tx.update(this.tx, {isPaperWallet: true})
+            this._handleScan(wallet.publicAddr)
+          })
+          .catch((err) => {
+            console.log('[ERROR]: Failed to print the paper wallet.', err)
+            this._idle()
+          })
+      })
+      .catch((err) => {
+        console.log('[ERROR]: The kiosk printer is in an invalid state.', err)
```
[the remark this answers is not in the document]
littlenaut: Yep, okay.

lib/kioskPrinter.js
```diff
@@ -0,0 +1,115 @@
...
+
+function printWalletZebra (wallet, printerCfg) {
+
+  return new Promise((resolve, reject) => {
+    const port = new SerialPort(printerCfg.address, zebraRS232Settings)
+    port.open((err) => {
+      if (err)
+        return reject(err)
+
+      const printWalletCmd = '^XA\n' +
```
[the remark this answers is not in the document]
littlenaut: This isn't a multiline string. It's a sequence of commands where some of them happen to have a line feed: `\n`.

lib/kioskPrinter.js
```diff
@@ -0,0 +1,115 @@
...
+  xoff: true
+}
+
+function checkStatus (printerCfg) {
+  return new Promise((resolve, reject) => {
+    if (printerCfg.maker === 'Zebra' &&
+        printerCfg.model === 'KR-403' &&
+        printerCfg.protocol === 'RS232' &&
+        printerCfg.address) {
+      return checkStatusZebra(printerCfg)
```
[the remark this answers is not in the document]
littlenaut: But within a month or so, there will be a `checkStatusNippon(...)` no?

lib/kioskPrinter.js
```diff
@@ -0,0 +1,115 @@
...
+      if (err)
+        return reject(err)
+
+      const getStatusCmd = '~HQES'
+      port.write(getStatusCmd, 'ascii', (err) => {
+        if (err)
+          return reject(err)
+
+        // Giving the printer 1sec to respond to the query about its status
+        setTimeout(() => {
```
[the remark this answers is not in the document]
littlenaut: I know that handling the `'data'` event will automatically put the port in flowing mode. I thought that was reserved for `RTS/CTS` connections but now I'm unsure. I'll have to confirm that software flow mode can operate in this scenario too.

lib/coins/btc.js
```diff
@@ -53,3 +54,13 @@ function validate (network, address) {
...
   return false
 }
+
+function createWallet () {
+    const keyPair = bitcoin.ECPair.makeRandom()
+    const segwitAddr = bitcoin.payments.p2wpkh({ pubkey: keyPair.publicKey })
+
+    return {
+        publicAddr: segwitAddr.address,
+        privateAddr: keyPair.toWIF()
```
joshmh: 👍

joshmh: 1. This is actually a privateKey not an address.
2. I think we should standardize on full names like publicAddress, otherwise we can easily forget the name of the variable which leads to bugs.

lib/kioskPrinter.js
```diff
@@ -29,26 +29,12 @@ function checkStatus (printerCfg) {
...
-          return reject(err)
-
-        // Giving the printer 1sec to respond to the query about its status
-        setTimeout(() => {
-          const printerStatusBuffer = port.read()
-          port.close()
-          return resolve(printerStatusBuffer)
-        }, 1000)
+      port.open((err) => {
+          if (err) reject(err)
```
joshmh: Best to use ``if (err) return reject(err)``, then we won't need the ``else``.

lib/kioskPrinter.js
```diff
@@ -66,50 +52,75 @@ function printWallet (wallet, printerCfg) {
   })
 }
 
+function checkStatusZebra (printerCfg) {
+  return new Promise((resolve, reject) => {
```
joshmh: Don't need to create a new Promise here since openSerialPort creates a promise we can chain off of.

lib/kioskPrinter.js
```diff
@@ -66,50 +52,75 @@ function printWallet (wallet, printerCfg) {
...
+    const port = new SerialPort(printerCfg.address, zebraRS232Settings)
+    openSerialPort(port)
+      .then(() => {
+        const getStatusCmd = '~HQES'
+        port.write(getStatusCmd, 'ascii', (err) => {
+          if (err)
+            return reject(err)
```
joshmh: Throw Error here.

lib/kioskPrinter.js
```diff
@@ -66,50 +52,75 @@ function printWallet (wallet, printerCfg) {
...
+                                   `^AQN,28,24^FD${wallet.publicAddr}^FS\n` +
+                               '^FO80,1170' +
+                                   `^BQN,2,10,H^FDHM,B0034${wallet.publicAddr}^FS\n` +
 
-                             '^CN1\n' +
-                             '^PN0\n' +
-                             '^XZ'
+                               '^CN1\n' +
+                               '^PN0\n' +
+                               '^XZ'
```
joshmh: I'm still not a fan of the + stuff. How about a list of commands that are then joined?

littlenaut: I think that's in fact the optimal and sane solution but only when we get into printing the receipts in which we must control the layout. Then we would have a set of functions, each with a very well defined responsibility (printing QRCode, printing logo, printing title, ...) that would be composed to assemble the command that then must be send to the printer.

lib/kioskPrinter.js
littlenaut: Can you elaborate on your idea please? I don't see how we could chain the events given that we'll need to wait on the `'data'` event to give us the printer status and then call `resolve` to continue the execution on the calling functions.

lib/kioskPrinter.js
joshmh: :+1:

lib/brain.js
```diff
@@ -1757,7 +1758,44 @@ Brain.prototype._start = function _start () {
...
+
+    this._transitionState('cashInWaiting')
+
+    const statusQueryTimeoutMs = 2000
+    kioskPrinter.checkStatus(deviceConfig.kioskPrinter, statusQueryTimeoutMs)
+      .then((printerStatus) => {
+        console.log('Kiosk printer status: ', printerStatus)
+        if (printerStatus.hasErrors) {
+          console.log('[ERROR]: The kiosk printer is in an invalid state.', err)
+          this._idle()
```
joshmh: Need an error screen here, too.

lib/brain.js
```diff
@@ -1757,7 +1758,44 @@ Brain.prototype._start = function _start () {
     return this.smsCompliance()
   }
 
+  const printPaperWallet = deviceConfig.compliance.paperWallet &&
+                           deviceConfig.kioskPrinter &&
+                           this.tx.cryptoCode === 'BTC'
```
joshmh: What happens if the coin isn't BTC?

lib/brain.js
```diff
@@ -1757,7 +1758,44 @@ Brain.prototype._start = function _start () {
...
+        }
+
+        const wallet = coinUtils.createWallet(this.tx.cryptoCode)
+        const printerCfg = deviceConfig.kioskPrinter
+        kioskPrinter.printWallet(wallet, printerCfg)
+          .then(() => {
+            this.tx = Tx.update(this.tx, {isPaperWallet: true})
+            this._handleScan(wallet.publicAddress)
+          })
+          .catch((err) => {
```
joshmh: Can this ``catch`` be combined with the next one? Perhaps we can just leave it out and let the outer ``catch`` handle all exceptions.

lib/coins/utils.js
```diff
@@ -67,3 +67,8 @@ function formatAddress (cryptoCode, address) {
   if (!plugin.formatAddress) return address
   return plugin.formatAddress(address)
 }
+
+function createWallet (cryptoCode) {
```
joshmh: Throw an error if not supported by the coin.

lib/kioskPrinter.js
```diff
@@ -0,0 +1,241 @@
...
+        .then((printerStatus) => resolve(printerStatus))
+        .catch((err) => reject(err))
+
+    return reject(errorMsgUnsupportedPrinter())
+  })
+}
+
+
+function printWallet (wallet, printerCfg) {
+  return new Promise((resolve, reject) => {
```
joshmh: We don't need to create our own Promise here because we're going to be calling functions that already return promises (e.g., ``printWalletZebra()``).

lib/kioskPrinter.js
```diff
@@ -0,0 +1,241 @@
...
+    const model = checkPrinterModel(printerCfg)
+
+    if (model === printerModels.zebra)
+      return printWalletZebra(wallet, printerCfg)
+        .then(() => resolve())
```
joshmh: Don't need ``then/catch`` just return the promise:

```
return printWalletZebra(wallet, printerCfg)
```

lib/kioskPrinter.js
```diff
@@ -0,0 +1,241 @@
...
+        .catch((err) => reject(err))
+
+    return reject(errorMsgUnsupportedPrinter())
```
joshmh: This can be replaced with:

```
return Promise.reject(errorMsgUnsupportedPrinter())
```

lib/kioskPrinter.js
joshmh: Yeah, since you're monitoring ``data`` in the end it's not that simple. I think we can keep it this way.

RafaelTaranto:
Superseded by https://github.com/lamassu/lamassu-machine/pull/296

lib/brain.js
RafaelTaranto: I'm moving that to the other PR

joshmh:
Included in https://github.com/lamassu/lamassu-machine/pull/296.

PR #256, opened 2019-03-11
Only present WiFi selection screen on Douro I models

dangerware:
This fix introduces a new requirement for `device_config.json` data structure. A new top-level field called `cryptomatModel`. For this specific fix I started by assuming there's a model called `Douro I`. The existence of this field will probably benefit other machine-side features.

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -216,8 +216,12 @@ Brain.prototype.run = function run () {
 
   this.clientCert = pairing.getCert(this.certPath)
 
-  if (!this.clientCert) return this._transitionState('virgin')
-  this.checkWifiStatus()
+  if (!this.clientCert)
+    return this._transitionState('virgin')
+
+  if (deviceConfig.cryptomatModel === 'douro1')
+    this.checkWifiStatus()
+  else this.initTrader()
 }
 
 Brain.prototype._executeCallbackAfterASufficientIdlePeriod =
@@ -288,7 +292,9 @@ Brain.prototype.initialize = function initialize () {
   pairing.init(this.certPath)
     .then(clientCert => {
       this.clientCert = clientCert
-      this.checkWifiStatus()
+      if (deviceConfig.cryptomatModel === 'douro1')
+        this.checkWifiStatus()
+      else this.initTrader()
     })
 }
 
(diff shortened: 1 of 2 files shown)


lib/brain.js
```diff
@@ -298,18 +298,27 @@ Brain.prototype.checkWifiStatus = function checkWifiStatus () {
...
+            self.wifi.startScanning()
+            self._wifiList()
+            return
+          }
+          self.config.ip = ip
+          self._wifiConnected()
+        })
+      }
+      else {
+        self.initTrader()
```
littlenaut: Note that the communication loop handled by the Trader module is what ends up presenting the 'Network Down' screen when too much negotiations with the server fail sequentially. It is also handles the recovery of the connection when it is lost.

lib/brain.js
```diff
@@ -298,18 +298,27 @@ Brain.prototype.checkWifiStatus = function checkWifiStatus () {
...
-        if (err) {
-          self.wifi.startScanning()
-          self._wifiList()
-          return
-        }
-        self.config.ip = ip
-        self._wifiConnected()
-      })
+    if (err) {
+      if (err)
```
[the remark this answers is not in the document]
joshmh: Explain what caused the bug and how the code changes fix it.

littlenaut:
The reported bug stated that the 'Select WiFi' screen was appearing in cryptomats that don't have an WiFi connection, like the Sintra or Douro II.
This was happening because this screen was being presented based on the feedback from the [`wifi.js`](https://github.com/lamassu/lamassu-machine/blob/dev/lib/wifi.js) module, which doesn't tells apart one from the other.

While we don't have the testing framework in place, I decided that the safest way to fix this was to verify the machine model once we receive the answer from wifi.js. If it is a Douro I, we go the usual WiFi screen, otherwise relay control to the trader which will recover automatically if there's an Ethernet connection or go to the 'Network Down' screen if there isn't one at all.

lib/brain.js
```diff
@@ -216,8 +216,12 @@ Brain.prototype.run = function run () {
 
   this.clientCert = pairing.getCert(this.certPath)
 
-  if (!this.clientCert) return this._transitionState('virgin')
-  this.checkWifiStatus()
+  if (!this.clientCert)
+    return this._transitionState('virgin')
+
+  if (deviceConfig.cryptomatModel === 'douro1')
+    this.checkWifiStatus()
```
joshmh: Use ``return`` instead of if/else.

lib/brain.js
```diff
@@ -298,9 +304,11 @@ Brain.prototype.checkWifiStatus = function checkWifiStatus () {
   this._transitionState('booting')
 
   this.wifi.status(function (err, status, ip) {
-    if (err || status === 'bill') {
-      if (err) console.log(err.stack)
-      if (self.state !== 'wifiConnecting') self._wifiConnecting()
+    if (err) {
```
joshmh: What about handling ``status === 'bill'``? If this is not actually necessary, explain why.

RafaelTaranto:
Superseded by https://github.com/lamassu/lamassu-machine/pull/286

lib/brain.js
RafaelTaranto: status === 'bill' was actually redundant since quite some time, the linked commit mistakenly replaced 'pending' with 'bill' in this line

https://github.com/lamassu/lamassu-machine/commit/54b0085f351182f53174751a533d96f3a3640b6c#diff-aa306c9a688d1483cc340067b667e2ffR214

PR #257, opened 2019-03-11
Prevent crash when trying to disable iVizon before it's fully initialised

dangerware:


littlenaut:
[Sintra iVizion boot requires door to be latched](https://trello.com/c/6N159g9p/133-sintra-ivizion-boot-requires-door-to-be-latched)

lib/id003/id003.js
```diff
@@ -175,6 +175,7 @@ Id003.prototype._run = function _run (cb) {
       }
     }, 500)
   })
+  this.initialized = true
```
joshmh: BillValidator isn't initialized yet, this should be set inside the callbacks after the validator gives us the denominations.

@littlenaut can you explain why you set this in parallel to opening the rs-232 port? This kind of thing can cause a lot of problems in nodeJS programming.

littlenaut: @joshmh You're right in should be placed on the line above.

littlenaut:
This PR fixes a crash that would occur at l-m startup if no connection to the server is available. After several failed attempts, the machine will jump into the `Network Down` state and try to disable the bill validator, which was never initialised to begin with because it requires a first successful connection with the server to do so.

By initialisation I mean that the JavaScript structure that deals with the bill validator wasn't fully initialised and the process would crash [here](https://github.com/lamassu/lamassu-machine/blob/42519b63f79f6dbc757a97f826d676cfdf2512d5/lib/id003/id003.js#L208) because `id003Fsm` was undefined.

The fix makes use of the already existent `initialized` field to signal that the object is fully initialised so it's safe to be used by the clients.

lib/brain.js
```diff
@@ -2045,7 +2045,8 @@ Brain.prototype.enableBillValidator = function enableBillValidator () {
 
 Brain.prototype.disableBillValidator = function disableBillValidator () {
   emit('billValidatorOff')
-  this.billValidator.disable()
+  if (this.billValidator.initialized)
```
joshmh: Since this is a low level error, I think it would be better to handle it in the billValidator module. It would be better if we didn't have to ``check billValidator.initialized`` from outside the billValidator module.

One reason for this is that we can crash the app by calling ``this.billValidator.disable()`` anywhere in the code if we're not careful to check that it's initialized. If we put the logic in billValidator, we should never be able to crash the app by calling it.

lib/id003/id003.js
joshmh: We shouldn't be sending commands until we're finished fully initializing the peripheral, otherwise those commands could get in the way of a successful initialization. The more important point is to be careful about following NodeJS's callback-based execution and not setting flags in parallel because that leads to a lot of subtle bugs that can be hard to catch and debug.

lib/brain.js
littlenaut: Hmm, if we're to solve this from the inside how should we deal with functions like [`isCashOutBox()`](https://github.com/lamassu/lamassu-machine/blob/dev/lib/id003/id003.js#L51) which either return `true` or `false` to the calling code? The "proper" way would be to throw an exception because the code that evaluates that hasn't been initialized, but in that case we also have to update the client code to deal with such cases.

joshmh: Yeah, an exception makes sense. An exception will be thrown anyway by nodejs, we'd just be making it more clear what the error is by catching and re-throwing. The important part is that higher level functions shouldn't have to worry about internal state of the functions they're calling.

RafaelTaranto:
Superseded by https://github.com/lamassu/lamassu-machine/pull/287

PR #260, opened 2019-03-15
Remove i18n duplicates and fix some messages

RafaelTaranto:


RafaelTaranto:
[Non-interpolation and duplicate machine UI strings](https://trello.com/c/2CLYTZRf/138-non-interpolation-and-duplicate-machine-ui-strings)

PR #263, opened 2019-03-25
Disabled choose fiat buttons style

RafaelTaranto:


RafaelTaranto:
[Buttons on choose fiat screen are not disabling](https://trello.com/c/rIIjqzNP/157-buttons-on-choose-fiat-screen-are-not-disabling)

PR #267, opened 2019-04-01
Periodically truncate logs

nurotyx:
At every synchronization with lamassu-server, the machine will check if there are any log files that are two or more days old and delete those.

### DIFF
--- a/lib/logs.js
+++ b/lib/logs.js
@@ -22,7 +22,7 @@ const MAX_LINES = 2000
  * @function
  * @async
  *
- * @param {date} timetamp Requested date as unix timestamp
+ * @param {date} timestamp Requested date as unix timestamp
  *
  * @returns {array} Array of log filenames that exceed the timetamp provided
  */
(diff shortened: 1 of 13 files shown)


nurotyx:
[Periodically truncate lamassu-machine files](https://trello.com/c/0ZtnDp9w/149-periodically-truncate-lamassu-machine-files)

lib/trader.js
```diff
@@ -62,6 +62,11 @@ Trader.prototype.syncLogs = function syncLogs () {
   // Get last seen timestamp from server
   this.request({ path: '/logs', method: 'get' })
     .then(data => data.body)
+  // Delete log files that are two or more days old
+    .then(() => {
+      let today = new Date()
+      return logs.removeLogFiles(today.setDate(today.getDate() - 2))
```
joshmh: I don't think this will work. ``getDate()`` gets the day of the month, not a date object. It doesn't seem like this code was tested.

nurotyx: Hmm, have you tried it in your computer? It's true that `getDate()` returns the day of the month which is fed to `setDate()` (effectively setting the date to two days ago) which in turn returns a Unix timestamp.

nodejs-date

lib/logs.js
```diff
@@ -33,6 +33,37 @@ function getNewestLogFiles (timestamp) {
...
+    const time = timestamp.slice(0, 10)
+    return _.filter(filename => filename.slice(0, 10) <= time, files)
+  })
+}
+
+function removeLogFiles (timestamp) {
+  const timestampISO = timestamp.toISOString()
+  return getOlderLogFiles(timestampISO)
+    .then(files => {
+      return Promisse.all(_.map(filename => {
```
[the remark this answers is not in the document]
nurotyx: I caught that error and another two or so around that line but they were left of staging, sorry and thank you for noticing. BTW all this code was tested on the ssuboard itself with log files generated by previous runs of l-m.

lib/trader.js
```diff
@@ -62,6 +62,11 @@ Trader.prototype.syncLogs = function syncLogs () {
   // Get last seen timestamp from server
   this.request({ path: '/logs', method: 'get' })
     .then(data => data.body)
+  // Delete log files that are two or more days old
+    .then(() => {
```
RafaelTaranto: data.body was passing through to logs.queryNewestLogs and now we return whatever logs.removeLogFiles return.

PR #270, opened 2019-04-02
Add KRW, correct ILS

naconner:


naconner:
[Add KRW, correct ILS on F56 bills](https://trello.com/c/SDD7gVr1/168-add-krw-correct-ils-on-f56-bills)

PR #271, opened 2019-04-03
Update reboot script to handle ssuboard systems

dangerware:
This reboot script will now handle ssuboard's lamassu-machine and lamassu-browser reboot through `supervirsorctl`.
Also added a readability improvement, regarding date handling, that was missing from #267.

### DIFF
--- a/lib/trader.js
+++ b/lib/trader.js
@@ -64,8 +64,17 @@ Trader.prototype.syncLogs = function syncLogs () {
     .then(data => data.body)
   // Delete log files that are two or more days old
     .then(() => {
-      let today = new Date()
-      return logs.removeLogFiles(today.setDate(today.getDate() - 2))
+      const twoDaysAgo = (() => {
+        let date = new Date()
+
+	// Notice that `setDate()` can take negative values. So if you'd take
+	// -2 days on the 1st of April you'd get the 30th of March. Several
+	// examples can be seen at: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate#Using_setDate()
+        date.setDate(date.getDate() - 2)
+	return date
+      })()
+
+      return logs.removeLogFiles(twoDaysAgo)
     })
   // Load unseen logs to send
     .then(logs.queryNewestLogs)
--- a/lib/trader.js
+++ b/lib/trader.js
@@ -74,7 +74,7 @@ Trader.prototype.syncLogs = function syncLogs () {
 	return date
       })()
 
-      return logs.removeLogFiles(twoDaysAgo).then(it)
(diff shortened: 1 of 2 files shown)


nurotyx:
[reboot update.tar for ssuboards](https://trello.com/c/Sr9zFj3U/153-reboot-updatetar-for-ssuboards)

RafaelTaranto:
@dangerware Can you passthrough the previous value on the promise here? I'll add comments on the PR on how to do this.

lib/trader.js
```diff
@@ -64,8 +64,17 @@ Trader.prototype.syncLogs = function syncLogs () {
     .then(data => data.body)
   // Delete log files that are two or more days old
     .then(() => {
```
RafaelTaranto: We need to pass data.body to logs.queryNewestLogs, name the param here.

```
.then((it) => {
```

lib/trader.js
```diff
@@ -64,8 +64,17 @@ Trader.prototype.syncLogs = function syncLogs () {
...
+        let date = new Date()
+
+	// Notice that `setDate()` can take negative values. So if you'd take
+	// -2 days on the 1st of April you'd get the 30th of March. Several
+	// examples can be seen at: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate#Using_setDate()
+        date.setDate(date.getDate() - 2)
+	return date
+      })()
+
+      return logs.removeLogFiles(twoDaysAgo)
```
RafaelTaranto: now return the param, that way this function runs without interfering with queryNewestLogs

```
return it
```

lib/trader.js
RafaelTaranto: Note that you can also refactor this delete stuff to a separate function and call it outside this promise chain, just make sure to add error handling to it.

lib/trader.js
```diff
@@ -63,9 +63,18 @@ Trader.prototype.syncLogs = function syncLogs () {
...
+        let date = new Date()
+
+	// Notice that `setDate()` can take negative values. So if you'd take
+	// -2 days on the 1st of April you'd get the 30th of March. Several
+	// examples can be seen at: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/setDate#Using_setDate()
+        date.setDate(date.getDate() - 2)
+	return date
+      })()
+
+      return logs.removeLogFiles(twoDaysAgo).then(it)
```
RafaelTaranto: `then()` expects a function as argument. This should be `then(() => it)`

joshmh: Thanks for the review, Rafi!

RafaelTaranto: It was mainly regarding the bug, though. Didn't look deep into the PR.

PR #273, opened 2019-04-09
Terms conditions pagination

RafaelTaranto:


RafaelTaranto:
[Terms & Conditions error rendering when in presence of Markdown or Emojis](https://trello.com/c/3S2TYcC9/172-terms-conditions-error-rendering-when-in-presence-of-markdown-or-emojis)

PR #274, opened 2019-04-10
Identify F56 error code in message log

dangerware:
This should now propagate the `statusCode` and properly log the F56 protocol error code. According to the F56 D-LEVEL spec, the error code lives at bytes 3 and 4 in the response frame.


dangerware:
[F53 dispenser errors not logged](https://trello.com/c/OuSSxjtt/144-f53-dispenser-errors-not-logged)

PR #275, opened 2019-04-12
On batch cash-out, wait for user to collect the notes before delivering next batches

dangerware:
The changes introduced here will wait for the use to collect a batch of notes before delivering the next one. Previously, all batches would be delivered one after the other without waiting for the user to collect them.


dangerware:
[F53 batch dispensing not respecting presence of batched note stack](https://trello.com/c/714VYnmu/145-f53-batch-dispensing-not-respecting-presence-of-batched-note-stack)

joshmh:
Incorporated in https://github.com/lamassu/lamassu-machine/pull/317.

PR #276, opened 2019-04-16
Fix the displayed cash-in rate

RafaelTaranto:


RafaelTaranto:
[Ensure correct commission rate display on cash-in](https://trello.com/c/pTRshihe/182-ensure-correct-commission-rate-display-on-cash-in)

PR #277, opened 2019-04-16
Optimize background img

RafaelTaranto:


RafaelTaranto:
[Choppy screen transitions on Chromium, lack of animations](https://trello.com/c/3TljGAk7/180-choppy-screen-transitions-on-chromium-lack-of-animations)

PR #279, opened 2019-04-17
Add new unseen tx screen

RafaelTaranto:


RafaelTaranto:
[User is told transaction is seen on the blockchain, even if it doesn't exist + grammar mistake](https://trello.com/c/bdOjYUGP/166-user-is-told-transaction-is-seen-on-the-blockchain-even-if-it-doesnt-exist-grammar-mistake)

PR #280, opened 2019-04-18
Stop appending to machine.log and remove old log files

RafaelTaranto:


RafaelTaranto:
[Restrict supervisor logs to prevent full disk](https://trello.com/c/HIhLpjyO/184-restrict-supervisor-logs-to-prevent-full-disk)

PR #284, opened 2019-04-23
Watch resize event for background adjustment

RafaelTaranto:


RafaelTaranto:
[UI bottom bar cut off when maximised on Aaeon](https://trello.com/c/23LwWjXA/189-ui-bottom-bar-cut-off-when-maximised-on-aaeon)

PR #291, opened 2019-05-02
Fix scan button center on surf/chromium

RafaelTaranto:


RafaelTaranto:
[Fix scan button text align](https://trello.com/c/RJkLZGQw/196-fix-scan-button-text-align)

PR #294, opened 2019-05-10
Create eth down screen for machines without wifi

RafaelTaranto:


RafaelTaranto:
['Select Wi-Fi' screen appears on ssuboard models when no Ethernet detected](https://trello.com/c/oTcXlWoo/53-select-wi-fi-screen-appears-on-ssuboard-models-when-no-ethernet-detected)

PR #296, opened 2019-05-13
WIP: Paper wallet printing

RafaelTaranto:
Supersedes https://github.com/lamassu/lamassu-machine/pull/254

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -1792,7 +1792,8 @@ Brain.prototype._start = function _start () {
 
     this._transitionState('cashInWaiting')
 
-    kioskPrinter.checkStatus(deviceConfig.kioskPrinter)
+    const statusQueryTimeoutMs = 2000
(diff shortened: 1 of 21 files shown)


RafaelTaranto:
[Private key printing](https://trello.com/c/A9y5mQE1/136-private-key-printing)

lib/brain.js
```diff
@@ -1784,10 +1793,73 @@ Brain.prototype._start = function _start () {
     return this.smsCompliance()
   }
 
+  const printPaperWallet = deviceConfig.compliance.paperWallet &&
+                           deviceConfig.kioskPrinter &&
+                           this.tx.cryptoCode === 'BTC'
```
RafaelTaranto: > What happens if the coin isn't BTC?

@joshmh what should happen if the coin isn't BTC? Do we need a screen for that 'misconfiguration'?

joshmh: The operator shouldn't allow unsupported currencies, so no need for this. We can just send the user back to idle.

PR #297, opened 2019-05-14
Make embedded svg hide cursor

RafaelTaranto:


RafaelTaranto:
[Tapping on animations allows mouse pointer to show, disappearing shortly afterwards](https://trello.com/c/Ok7yjLkS/211-tapping-on-animations-allows-mouse-pointer-to-show-disappearing-shortly-afterwards)

joshmh:
@RafaelTaranto Is it possible to style this with just a single CSS descriptor in the main file instead of needing to add that descriptor in every SVG file? Or is it possible to hide the cursor at the OS level just for the browser window?

RafaelTaranto:
> Is it possible to style this with just a single CSS descriptor in the main file instead of needing to add that descriptor in every SVG file?

Not really, we're using Object on the animated svgs because of the older chromium version and that starts a new document.

> Or is it possible to hide the cursor at the OS level just for the browser window?

I'll do some research into this.

joshmh:
Ok, let's keep it like this.

PR #298, opened 2019-05-14
Fix rtl close button

RafaelTaranto:


RafaelTaranto:
[Cancel button in middle of screen on RTL languages](https://trello.com/c/euyAmJ7X/208-cancel-button-in-middle-of-screen-on-rtl-languages)

PR #300, opened 2019-05-14
Small printer fix and svgs

RafaelTaranto:


RafaelTaranto:
[Cannot read property 'paperWallet' of undefined](https://trello.com/c/8huL7cp5/205-cannot-read-property-paperwallet-of-undefined)

PR #301, opened 2019-05-15
Add TWD to f56 bills

naconner:


naconner:
[Add TWD to F56 bills](https://trello.com/c/Vzl7gjdz/216-add-twd-to-f56-bills)

PR #303, opened 2019-05-18
New id verification screen

RafaelTaranto:


RafaelTaranto:
[ID request references phone number, instead of ID card](https://trello.com/c/5ijNFtbj/199-id-request-references-phone-number-instead-of-id-card)

ui/js/keypad.js
```diff
@@ -25,9 +25,10 @@ var Keypad = function (keypadId, opts, callback) {
 
     if (target.hasClass('enter')) {
       self.deactivate()
-      var result = self.type === 'phoneNumber'
-        ? libphonenumber.parsePhoneNumberFromString(self.result, self.opts.country).number
-        : self.result
+      var result = self.result
+      if (self.type === 'phoneNumber' && libphonenumber.parsePhoneNumberFromString(self.result, self.opts.country)) {
```
joshmh: We should only run parsePhoneNumberFromString once, and then check the variable to make sure it worked. Also, can result be a const?

ui/src/app.js
```diff
@@ -250,7 +250,11 @@ function processData (data) {
       chooseCoin(data.coins, data.twoWayMode)
       break
     case 'smsVerification':
-      smsVerification(data.threshold)
+      if (data.hasPhone) {
+        idVerification()
```
joshmh: Why are we going to idVerification in the case of smsVerification? Shouldn't we have an idVerification case?

PR #304, opened 2019-05-21
Don't log sync logs connection errors

RafaelTaranto:


RafaelTaranto:
[Don't pollute l-m logs with log sync error when network is down](https://trello.com/c/ABVROvEV/214-dont-pollute-l-m-logs-with-log-sync-error-when-network-is-down)

PR #305, opened 2019-05-21
Scan PK on printer flow

RafaelTaranto:


RafaelTaranto:
[Private key printing](https://trello.com/c/A9y5mQE1/136-private-key-printing)

lib/scanner.js
```diff
@@ -171,6 +172,20 @@ function scanMainQR (cryptoCode, callback) {
...
 }
 
+function scanPK (callback) {
+  scanQR(function (err, result) {
+    if (err) return callback(err)
+    if (!result) return callback(null, null)
+
+    var resultStr = result.toString()
+    try {
+      callback(null, resultStr)
```
joshmh: Why do we want to ``try`` the callback? Shouldn't we be ``try``ing ``toString``?

joshmh:
Can this PR be closed since it's superseded by #306 ?

PR #309, opened 2019-05-27
Update report.js for ssuboard / upboard

naconner:


naconner:
[reboot update.tar for ssuboards](https://trello.com/c/Sr9zFj3U/153-reboot-updatetar-for-ssuboards)

PR #315, opened 2019-05-28
Remove hardcoded reason from sendOnly state

RafaelTaranto:


RafaelTaranto:
[New 'out of coins' screen refers to daily limit when total balance is reached](https://trello.com/c/wBh2nj2X/227-new-out-of-coins-screen-refers-to-daily-limit-when-total-balance-is-reached)

PR #316, opened 2019-05-28
Add test script for the Nippon's kiosk printer

dangerware:
This test script should print a welcoming message and a QR code for https://lamassu.is, using the serial interface for the Nippon NP-2511D-2 kiosk printer.

It should be executed like `sudo node nipponPrinterTests.js`.  
It can take two parameters: `--dev` and `--str`. The first will select a different serial port than the default one (`/dev/ttyJ5`) and the second let the user set the content of the QR code.

### DIFF
--- a/tools/nipponPrinterTests.js
+++ b/tools/nipponPrinterTests.js
@@ -1,0 +1,52 @@
+const minimist = require('minimist')
+const SerialPort = require('serialport')
+
+const portOptions = {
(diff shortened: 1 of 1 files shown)


dangerware:
[Setup and test Gaia's Kiosk Printer](https://trello.com/c/bF6C3vWm/35-setup-and-test-gaias-kiosk-printer)

PR #317, opened 2019-05-29
F56 batch dispense

RafaelTaranto:


RafaelTaranto:
[F53 batch dispensing not respecting presence of batched note stack](https://trello.com/c/714VYnmu/145-f53-batch-dispensing-not-respecting-presence-of-batched-note-stack)

PR #318, opened 2019-05-31
Fix calculation to display hardLimitReached

RafaelTaranto:


RafaelTaranto:
[No message given for returning customer with previously maxed daily limit](https://trello.com/c/uW6Je8Vw/220-no-message-given-for-returning-customer-with-previously-maxed-daily-limit)

PR #319, opened 2019-05-31
Store operator info locally

RafaelTaranto:


RafaelTaranto:
[Store operator contact info locally, persist through OS reboots](https://trello.com/c/fDSKktoS/229-store-operator-contact-info-locally-persist-through-os-reboots)

PR #321, opened 2019-06-02
Identify F56 error code in message log (reprise)

dangerware:
This is a continuation of #274.
This objective now is to also propagate the report to the error log that is sent to the server.


dangerware:
[F53 dispenser errors not logged](https://trello.com/c/OuSSxjtt/144-f53-dispenser-errors-not-logged)

PR #324, opened 2019-06-13
Stop polling when doing another request on id003

RafaelTaranto:


RafaelTaranto:
[Fix id003 bug (CRC error / badFrame)](https://trello.com/c/V81L5HFk/241-fix-id003-bug-crc-error-badframe)

lib/id003/id003.js
```diff
@@ -240,11 +238,25 @@ Id003.prototype._denominations = function _denominations () {
...
 
 Id003.prototype._poll = function _poll () {
-  this._send('status')
+  this.rs232.send('status')
 }
 
 Id003.prototype._send = function _send (command, data) {
-  this.rs232.send(command, data)
+  this._stopPolling()
+  setTimeout(() => {
```
joshmh: Is this timeout necessary, since _startPolling uses and interval anyway and won't start immediately?

RafaelTaranto: Ahh, good point!

PR #325, opened 2019-06-13
CCnet implementation

RafaelTaranto:


RafaelTaranto:
[Gaia - CPI MSM CCNET protocol support](https://trello.com/c/jUg2GIOh/185-gaia-cpi-msm-ccnet-protocol-support)

lib/ccnet/ccnet.js
```diff
@@ -0,0 +1,208 @@
...
+
+Ccnet.prototype.enable = function () {
+  const billsBuffer = this.getBillsToEnable()
+
+  const command = Buffer.concat([Buffer.from(commands.ENABLE_BILL_TYPES), billsBuffer])
+  this._send(command)
+}
+
+Ccnet.prototype.disable = function () {
+  const command = Buffer.concat([Buffer.from(commands.ENABLE_BILL_TYPES), Buffer.alloc(6)])
```
joshmh: Maybe *commands* should already be buffers, then we won't have to do `Buffer.from` each time.

lib/ccnet/ccnet.js
```diff
@@ -0,0 +1,208 @@
...
+
+Ccnet.prototype._handleResponse = function (response, command) {
+  if (command === commands.GET_BILL_TABLE) {
+    return this._billTable(response)
+  }
+
+  return statesFsm.handle(respMap[response[0]], response.slice(1), command)
+}
+
+Ccnet.prototype._billTable = function (data) {
```
joshmh: Since this creates side effects and isn't a pure function, describe side effects in a comment.

lib/ccnet/consts.js
```diff
@@ -0,0 +1,93 @@
+const SYNC = 0x02
+const ADDRESS = 0x03
+
+const commands = {
```
joshmh: Can make these all buffers by running the values through ``Buffer.from`` in lodash.

lib/id003/id003.js
```diff
@@ -243,8 +243,10 @@ Id003.prototype._poll = function _poll () {
 
 Id003.prototype._send = function _send (command, data) {
   this._stopPolling()
-  this.rs232.send(command, data)
-  this._startPolling()
+  setTimeout(() => {
```
joshmh: Why do we need this extra delay? Add a comment.

lib/ccnet/ccnet.js
RafaelTaranto: Is this for performance reasons or styling? I'm currently using commands without Buffer.from more often then not on code, but at the end rs232 code always redoes it.

joshmh: Just to make the code more compact. If all the commands are predefined as buffers it should make the code more simple.

PR #326, opened 2019-06-13
Stop scan pdf 417 from failing fast on missread and fix minimum fiat error on cash-in

RafaelTaranto:


RafaelTaranto:
[Camera 'already closed' during ID scan](https://trello.com/c/TiXK9Jg9/238-camera-already-closed-during-id-scan)

PR #328, opened 2019-06-16
Fix double goodbye screens after dispense

RafaelTaranto:


RafaelTaranto:
[goodbye screen shown twice, interrupts transaction in course](https://trello.com/c/5RlWVBVT/242-goodbye-screen-shown-twice-interrupts-transaction-in-course)

PR #332, opened 2019-06-19
Add new fonts

naconner:


naconner:
[Add new fonts](https://trello.com/c/eCDMHa9G/249-add-new-fonts)

PR #333, opened 2019-06-19
Extend scanner timeout

naconner:
This had been present in a DO version of this file used in the build process, but our repo's version should be used instead.


naconner:
[Extend scanner timeout](https://trello.com/c/P8OIyEFf/251-extend-scanner-timeout)

PR #334, opened 2019-06-20
Fix compliance amount calc

RafaelTaranto:


RafaelTaranto:
[Further cash-out compliance at first dollar causes issue](https://trello.com/c/BwiR9cg7/247-further-cash-out-compliance-at-first-dollar-causes-issue)

PR #335, opened 2019-06-22
Add GHS to MEI denominations

naconner:


naconner:
[Add GHS to MEI denominations](https://trello.com/c/ELs6K7nd/250-add-ghs-to-mei-denominations)

PR #337, opened 2019-06-23
deploy package changes

naconner:
We should preserve both the validator type (*i.e.*, `cashflowSc`) as well as its mount point during updates.


naconner:
[Deploy package changes](https://trello.com/c/fRW3AxLW/252-deploy-package-changes)

PR #338, opened 2019-06-24
Use copy-node-modules to ignore dev dependencies

RafaelTaranto:


RafaelTaranto:
[Reinstate `copy-node-modules` in codebase build script](https://trello.com/c/Ej69FzBl/254-reinstate-copy-node-modules-in-codebase-build-script)

PR #354, opened 2019-07-15
Remove bitcoincash: string from display address

RafaelTaranto:


RafaelTaranto:
[Remove "bitcoincash:" from address format and preserve in the QR code.](https://trello.com/c/J2vBrAoI/22-remove-bitcoincash-from-address-format-and-preserve-in-the-qr-code)

PR #355, opened 2019-07-16
Add new reboot option and restart lamassu-browser

RafaelTaranto:


RafaelTaranto:
[New admin reboot options](https://trello.com/c/8cZymyGC/6-new-admin-reboot-options)

PR #357, opened 2019-07-16
Correct screen text typo

naconner:


joshmh:
Doesn't this have to be fixed in the html file as well?

naconner:
@joshmh Thanks, change made.

PR #358, opened 2019-07-23
Add expiration date and document number to the db

RafaelTaranto:


RafaelTaranto:
[Parse and add 'id_card_data_number' and 'id_card_data_expiration' to 'customers' table via ID Card Data](https://trello.com/c/LaMi54Zz/14-parse-and-add-idcarddatanumber-and-idcarddataexpiration-to-customers-table-via-id-card-data)

PR #359, opened 2019-07-23
Preload language dropdown fonts

RafaelTaranto:


RafaelTaranto:
[Languages in menu load after delay.](https://trello.com/c/JecGWNEp/42-languages-in-menu-load-after-delay)

PR #361, opened 2019-07-23
Fix F56 handling of invalid denominations

RafaelTaranto:


RafaelTaranto:
[l-m F53 code doesn't handle invalid bills](https://trello.com/c/1AjiRPyq/15-l-m-f53-code-doesnt-handle-invalid-bills)

PR #362, opened 2019-07-23
Increase buttons touch size + fix footer overflow

RafaelTaranto:


RafaelTaranto:
[Increase button size where possible.](https://trello.com/c/RXcvoI4B/41-increase-button-size-where-possible)

PR #364, opened 2019-07-28
Add GHS for F56

naconner:


naconner:
[Add GHS for F56](https://trello.com/c/eSdFdZbp/54-add-ghs-for-f56)

PR #365, opened 2019-07-30
Force trader poll on idle

RafaelTaranto:


RafaelTaranto:
[Force wallet balance update before allowing subsequent transaction](https://trello.com/c/xK4GDIlm/21-force-wallet-balance-update-before-allowing-subsequent-transaction)

lib/brain.js
```diff
@@ -1341,10 +1342,16 @@ Brain.prototype._idle = function _idle (locale) {
 
   this.tx = Tx.update(this.tx, {fiatCode: this.fiatCode})
 
+  pollPromise
+    .then(() => this._idleByMode(self.localeInfo))
```
joshmh: Should be ``this.localeInfo`` or just ``localInfo``.

PR #368, opened 2019-08-02
Add version number to the screen

RafaelTaranto:


RafaelTaranto:
[Display machine version number in corner of initialise and pair screens](https://trello.com/c/D1Io5QN9/33-display-machine-version-number-in-corner-of-initialise-and-pair-screens)

PR #373, opened 2019-08-08
Preserve compliance in device_config upon update

naconner:


naconner:
[Preserve compliance in device_config upon update](https://trello.com/c/2VHt6Qza/63-preserve-compliance-in-deviceconfig-upon-update)

PR #379, opened 2019-08-09
add nan to devDependencies

naconner:


naconner:
[add nan to devDependencies](https://trello.com/c/rcoDVHWO/64-add-nan-to-devdependencies)

PR #381, opened 2019-08-09
Remove mock-cam from package build

naconner:
Otherwise, error as mock-cam.js had been removed.


naconner:
[Remove mock-cam from package build](https://trello.com/c/z7AGUsmk/67-remove-mock-cam-from-package-build)

PR #384, opened 2019-08-18
Mitigate some click delay issues

RafaelTaranto:


RafaelTaranto:
[Browser button behaviour](https://trello.com/c/7lwhwZ7J/68-browser-button-behaviour)

ui/src/app.js
```diff
@@ -722,6 +722,8 @@ function targetButton (element) {
 
 function touchEvent (element, callback) {
   function handler (e) {
+    // Only handle touch events on gaia
+    if (e.type === 'touchend' && cryptomatModel !== 'gaia') return
```
joshmh: I think this should be by browser version. We want touch events on all future models (Sintra basic and Sintra forte, which use upboard).

PR #387, opened 2019-09-09
Fix browser version regex

RafaelTaranto:


RafaelTaranto:
[Unmatched parentheses in app.js](https://trello.com/c/1yVajTx0/78-unmatched-parentheses-in-appjs)

PR #394, opened 2019-09-19
Final fix machine logs

RafaelTaranto:


RafaelTaranto:
[Unexpected token in logs results in 'clock skew', 'internal server error', rendering machines down](https://trello.com/c/yLrrKD0D/81-unexpected-token-in-logs-results-in-clock-skew-internal-server-error-rendering-machines-down)

PR #406, opened 2019-10-01
Get DD scripts up to date

RafaelTaranto:


hardware/system/upboard/install_script.sh
```diff
@@ -13,7 +13,7 @@ sudo -E bash setup_8.x.sh
 sudo apt install nodejs -y
 
 git clone https://github.com/lamassu/lamassu-led
-git clone https://github.com/lamassu/lamassu-machine -b crafty-chnemu
+git clone https://github.com/lamassu/lamassu-machine -b defiant-dingirma
```
naconner: It'd be better to then check out a version tag rather than leave it on the branch. Otherwise, there could be commits without a version bump, and the support team will reasonably assume the machine is on v7.4.1 if it reports as v7.4.1.

PR #407, opened 2019-10-01
Fixed a (probable) bug in tx limit check at sell.

zouppen:
Tx limit should be checked against the whole amount instead of a single bill. Let me know if this is not a bug. At least it looks like it, when txLimit is set, it just limits the bills available but doesn't limit the maximum amount. I.e. when txLimit is 40 and 20+50 euro bills are available, the 50€ button is disabled but user may click 20€ button any number of times, exceeding 40 euro.

This patch makes the limit behave so that the limit is imposed on the whole amount so that when txLimit is 40, maximum of 40€ may withdrawn.

### DIFF
--- a/lib/tx.js
+++ b/lib/tx.js
@@ -124,7 +124,7 @@ function computeCashOut (tx, cassettes, virtualCassettes, txLimit) {
   const denominationIsAvailable = denom =>
     !!BillMath.makeChange(cassettes, tx.fiat.add(denom))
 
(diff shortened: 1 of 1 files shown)


RafaelTaranto:
Hey @zouppen, thanks for the PR. Not really a bug because txLimit is updated for every button press.

zouppen:
That's quite misleading naming if it's so. It should be denominationLimit because it only limits bills larger than that amount and is not checked against the full amount. Have you verified yourself that it works correctly in case the txLimit has non-infinite value?

PR #408, opened 2019-10-01
Mock camera improvements

zouppen:
This PR has two related but distinct patches:

* Allow reading mock camera QR code data dynamically from a file or FIFO instead of static content. Makes it easier to debug UI state transformations because app is not needed to restart while testing many QR codes.
* Make mock camera behave more like a real camera by keeping the "opened" state consistent. Without this patch `isOpened()` returns always true, making some operations to fail on mock camera.

### DIFF
--- a/lib/mocks/scanner.js
+++ b/lib/mocks/scanner.js
@@ -1,6 +1,8 @@
 const _ = require('lodash/fp')
 const Pdf417Parser = require('../compliance/parsepdf417')
 const coinUtils = require('../coins/utils')
+const fs = require('fs')
+const path = require('path')
 
 module.exports = {
   config,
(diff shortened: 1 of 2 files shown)


lib/mocks/scanner.js
```diff
@@ -28,16 +31,24 @@ function config (_configuration) {
...
   }, cbTimeout)
 }
 
 function scanMainQR (cryptoCode, callback) {
+  opened = true
   _cancelCb = callback
   handle = setTimeout(function () {
-    var resultStr = mockData.qrData[cryptoCode]
+    opened = false
+    if (mockData.qrDataSource === undefined) {
```
RafaelTaranto: No need to compare it to undefined: 

`if (!mockData.qrDataSource) {`

lib/mocks/scanner.js
```diff
@@ -28,16 +31,24 @@ function config (_configuration) {
...
+      var resultStr = mockData.qrData[cryptoCode]
+    } else {
+      resultStr = fs.readFileSync(path.join(__dirname, "../../", mockData.qrDataSource))
```
RafaelTaranto: Styling change: needs to use single quotes

zouppen:
Thanks for the review, changes applied.

joshmh:
Thanks, @zouppen!

PR #410, opened 2019-10-02
Fix puloon dispenser code

RafaelTaranto:


RafaelTaranto:
['dispense error2 arr.reduce is not a function'](https://trello.com/c/GZCrfKNl/90-dispense-error2-arrreduce-is-not-a-function)

PR #411, opened 2019-10-03
Add support for Cardano cryptocurrency

detrandix:


lib/coins/ada.js
```diff
@@ -0,0 +1,41 @@
+const b58 = require('b58')
+const cbor = require('cbor-sync')
+const CRC32 = require('crc-32'); 
+
+module.exports = {depositUrl, parseUrl}
+
+function parseUrl(network, address) {
+  console.log('DEBUG16: [%s] *%s*', network, address)
```
rizip1: I suppose this should be removed.

lib/coins/ada.js
```diff
@@ -0,0 +1,41 @@
...
+
+  // Cardano in Byron era not recognize network type (main/test) by address prefixes like f.e. bitcoin
```
rizip1: Typo: `not` -> `does not`, `f.e.` -> `e.g.`

lib/coins/ada.js
```diff
@@ -0,0 +1,41 @@
...
+  if (!validate(address)) return null
```
rizip1: Convention: Lets move `validate` above this `parseUrl` function, so it is not used sooner than defined.

rizip1: This is how other currencies handle nonvalid case right? No need to throw some exception? (just asking)

lib/coins/ada.js
```diff
@@ -0,0 +1,41 @@
+const b58 = require('b58')
```
rizip1: This package has 8 downloads in a week so I am quite afraid of using it. However, I believe there should be something like that already available as other cryptocurrencies should need it. Could you please check current dependencies or how other currencies do this?

rizip1: It seems you can use already install "base-x" package.

lib/coins/ada.js
```diff
@@ -0,0 +1,41 @@
...
+
+  return address
+}
+
+function depositUrl(address, amount) {
+  return `${address}?amount=${amount}`
+}
+
+function validate(address) { 
+  const bytes = b58.decode(address)
```
rizip1: What happens if an address is not in base58? Would not it throw some exception?

test/unit/test-coin-parse-url.js
```diff
@@ -64,3 +65,27 @@ test('Should throw for invalid ZEC address', t => {
...
+    'Ae2tdPwUPEZKmwoy3AU3cXb5Chnasj6mvVNxV1H11997q3VW5ihbSfQwGpm' // Yoroi like address
+  ]
+  
+  for (let addr of addresses) {
+    let parsed = parseADAAddress('main', addr)
+    t.is(parsed, addr)
+  }
+})
+
+test('Should recognize invalid ADA address', t => {
```
rizip1: What about test where address is not base58 encoded?

rizip1: I would either rename it to "Should recognize invaliad ADA addresses" as you test multiple addresses, but rather would create a separate test for each "invalid" case.

test/unit/test-coin-parse-url.js
```diff
@@ -64,3 +65,27 @@ test('Should throw for invalid ZEC address', t => {
   } catch (error) { }
   t.true(spy.threw())
 })
+
+test('Should parse ADA address (Byron era)', t => {
```
rizip1: Also here it could be better to divide this into two standalone tests.

lib/coins/ada.js
detrandix: Yes, this is how non-valid cases are handled.

lib/coins/ada.js
detrandix: There is also a `bs58check` package, but that does BTC (and other similar currencies) specific CRC checking (Cardano have different algorithm), so from this point of view `base-x` looks good.

test/unit/test-coin-parse-url.js
detrandix: OK

lib/coins/ada.js
detrandix: I will add try/catch blocks and return false

test/unit/test-coin-parse-url.js
detrandix: OK, will do separate tests :)

lib/coins/ada.js
detrandix: This log is in all other coins (except Ethereum) - so are you sure about remove this?

lib/coins/ada.js
detrandix: You are right, they fire exceptions, but for BTC this is little hidden in bs58check library

lib/coins/ada.js
```diff
@@ -0,0 +1,50 @@
+const BASE58 = '123456789ABCDEFGHJKLMNPQRSTUVWXYZabcdefghijkmnopqrstuvwxyz'
```
rizip1: It seems you miss `I` here, is not there a better way to create that? Using some JS methods returning chars for ascii positions?

lib/coins/ada.js
```diff
@@ -0,0 +1,50 @@
...
+const b58 = require('base-x')(BASE58)
+const cbor = require('cbor-sync')
+const CRC32 = require('crc-32'); 
+
+module.exports = {depositUrl, parseUrl}
```
rizip1: Please move this at the bottom of file

lib/coins/ada.js
rizip1: Ok in that case leave it here

lib/coins/ada.js
detrandix: They use this in all other coins files, so I pick this as a convention, so do you think is a good idea to move that?

lib/coins/ada.js
detrandix: > Compared with Base64, the following similar-looking letters are omitted: 0 (zero), O (capital o), I (capital i) and l (lower case L) as well as the non-alphanumeric characters + (plus) and / (slash).

lib/coins/ada.js
rizip1: Ok then leave it at the top

lib/coins/ada.js
rizip1: sorry, my bad

lib/coins/ada.js
```diff
@@ -0,0 +1,50 @@
...
+  const addrData = cborData[0]
+  const addrChecksum = cborData[1]
+  
+  const checksum = CRC32.buf(addrData) >>> 0 // force usigned integer
+  
+  if (addrChecksum !== checksum) {
+    throw new Error(`Ivalid address ${address}: invalid checksum`)
+  }
+
+  return true
```
rizip1: This return statement now does not make much sense as there are exceptions

PR #414, opened 2019-10-14
fix: log on f53 and dispense on puloon

RafaelTaranto:


RafaelTaranto:
['dispense error2 arr.reduce is not a function'](https://trello.com/c/GZCrfKNl/90-dispense-error2-arrreduce-is-not-a-function)

lib/brain.js
```diff
@@ -3142,10 +3142,13 @@ Brain.prototype._dispense = function _dispense () {
...
-          cassettes,
-          fiatCode: this.trader.locale.fiatCode
-        })
+        // wait for port to close
+        setTimeout(() => {
+          return this.billDispenser.init({
+            cassettes,
+            fiatCode: this.trader.locale.fiatCode
+          })
+        }, 200)
```
joshmh: Can you explain in more detail why we need this delay and why specifically 200ms?

PR #419, opened 2019-10-18
Add BGN

GenericCPU:


naconner:
Thanks. We'll need this PR against defiant-dingirma rather than dev, though. Resubmit?

naconner:
Also, needs `polymer`.

PR #420, opened 2019-10-18
fix: match decimal points in exchange rates

lcsfelix:


joshmh:
[Operator choice of machine-displayed crypto units](https://trello.com/c/ytcgzrUz/22-operator-choice-of-machine-displayed-crypto-units)

naconner:
[Crypto price UI precision higher on cash-out than cash-in](https://trello.com/c/w9kFSPNN/59-crypto-price-ui-precision-higher-on-cash-out-than-cash-in)

ui/js/app.js
```diff
@@ -1183,29 +1183,24 @@ function splitNumber(localize, localeCode) {
...
 
 function formatFiat(amount, fractionDigits) {
   if (!fractionDigits) fractionDigits = 0;
 
-  var localized = amount.toLocaleString(jsLocaleCode, {
-    useGrouping: true,
-    maximumFractionDigits: fractionDigits,
-    minimumFractionDigits: fractionDigits
-  });
-  return splitNumber(localized, jsLocaleCode) + ' ' + fiatCode;
```
RafaelTaranto: This is used for the internationalization, we cannot remove it.

ui/js/app.js
```diff
@@ -1183,29 +1183,24 @@ function splitNumber(localize, localeCode) {
...
-    minimumFractionDigits: 3
+    maximumFractionDigits: fractionDigits,
+    minimumFractionDigits: fractionDigits
   });
 
   return splitNumber(localized, jsLocaleCode);
 }
 
-function formatCrypto(amount) {
+function formatCrypto(amount, fractionDigits) {
```
RafaelTaranto: This method is for formating crypto values, I don't think we need to change it

ui/js/app.js
```diff
@@ -1223,7 +1218,7 @@ function setExchangeRate(_rates) {
 
   if (rates.cashOut) {
     var cashOut = new BigNumber(rates.cashOut);
-    var cashOutCryptoToFiat = cashOut && formatCrypto(cashOut.round(3).toNumber());
+    var cashOutCryptoToFiat = cashOut && formatCrypto(cashOut.round(3).toNumber(), 2);
```
RafaelTaranto: See on line 1216 how we do it for the cash-in, we should the same here. I.E. replace formatCrypto with formatFiat and add the second parameter "2".

ui/js/app.js
```diff
@@ -1183,29 +1183,24 @@ function splitNumber(localize, localeCode) {
   return ['<span class="integer">', split[0], '</span><span class="decimal-char">', decimalChar, '</span><span class="decimal">', split[1], '</span>'].join('');
 }
 
-function formatNumber(num) {
+function formatNumber(num, fractionDigits = 3) {
```
RafaelTaranto: No changes needed on this function, since it's only used for formatting crypto.

ui/js/app.js
lcsfelix: Changing only this results still results in 3 decimal points, with a trailing zero, due to the way the formatNumber, which is called by formatCrypto, is written

PR #430, opened 2019-11-06
Fix resolution for front facing camera

GenericCPU:
1280x960 is not a supported resolution for this camera

### DIFF
--- a/hardware/codebase/ssuboard/device_config.json
+++ b/hardware/codebase/ssuboard/device_config.json
@@ -22,7 +22,7 @@
     "device": "/dev/video-front",
     "facephoto": {
       "width": 1280,
-      "height": 960,
+      "height": 1024,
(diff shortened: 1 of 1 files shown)


joshmh:
Doesn't this conflict with https://github.com/lamassu/lamassu-machine/pull/429/files? Also, there should be a trello card for this. Will this work with all cameras we shipped with the ssuboard?

GenericCPU:
The earlier pull request can be ignored, we decided to go wit 1024 instead of 720. This will work with the Sintra's camera model:
ELP 5MP: "ELP-USB500W05G-FD100 "

All ssuboard devices should be using this camera.

GenericCPU:
https://trello.com/c/2qsUSs4D/108-front-facing-camera-not-working-sintra

PR #431, opened 2019-11-14
Support for upboard on sintra

RafaelTaranto:


naconner:
[Support for upboard on Sintra](https://trello.com/c/u4FkKiKV/98-support-for-upboard-on-sintra)

hardware/codebase/upboard/sintra/device_config.json
```diff
@@ -0,0 +1,66 @@
...
+  "brain": {
+    "freeMemRatio": 0.01,
+    "powerStatus": null,
+    "dataPath": "data",
+    "wifiConfigPath": null,
+    "nfcReader": null,
+    "wifiDisabled": true
+  },
+  "kioskPrinter": {
+    "maker": "Zebra",
```
joshmh: This depends on the machine, some will use Nippon printers.

hardware/codebase/upboard/sintra/device_config.json
```diff
@@ -0,0 +1,66 @@
...
+    "address": "/dev/ttyJ4"
+  },
+  "compliance": {
+    "paperWallet": false
+  },
+  "frontFacingCamera": {
+    "device": "/dev/video-front",
+    "facephoto": {
+      "width": 1280,
+      "height": 960,
```
joshmh: According to cam specs: 1280×720@30fps MJPEG; 5fps YUY2

https://www.kayetoncctv.com/non-distortion-5-megapixel-manual-focus-uvc-otg-support-usb-camera-module-5mp-ov5640-webcam-for-linux-android-windows-mac/

hardware/system/upboard/install_script.sh
```diff
@@ -1,19 +1,32 @@
...
+  echo 'Install script expects "gaia" or "sintra" as machine parameter'
+  exit 1
+fi
+
 sudo apt update && sudo apt full-upgrade -y
 
 # install dependencies
 sudo apt install build-essential chromium curl git supervisor yasm -y
-sudo apt install libavcodec-dev libavformat-dev libswscale-dev -y
+sudo apt install libavcodec-dev libavformat-dev libswscale-dev libpcsclite-dev -y
```
joshmh: Why do we need ``libpcsclite-dev`` (or any of these)?

lib/upboard/sintra/board-manager.js
```diff
@@ -0,0 +1,13 @@
...
+const ledManager = require('../../ssuboard/led-manager')
+
+module.exports = { run }
+
+function run () {
+  const tasks = [
+    ledManager.run
+  ]
+
+  return pSeries(tasks)
```
joshmh: Why do we need pSeries if there's only one task?

watchdog.js
```diff
@@ -59,7 +60,7 @@ function executeScript () {
     return
   }
   console.log('in execute')
-  child = cp.fork(SCRIPT_PATH, [platform])
+  child = cp.fork(SCRIPT_PATH, [platform, model])
```
joshmh: Have to be careful here because old systems won't have this update until after the update. So the first update package will not be supplied the ``model`` param.

RafaelTaranto: right, model is being defaulted to gaia since every upboard+sintra will have this set

hardware/system/upboard/install_script.sh
RafaelTaranto: Good point, I think I can remove all audio related stuff from the code today

PR #433, opened 2019-11-20
fix: properly clear terms screen timeout

RafaelTaranto:


naconner:
[Terms and conditions timeouts too quickly after multiple tries](https://trello.com/c/tUjqzW5U/103-terms-and-conditions-timeouts-too-quickly-after-multiple-tries)

PR #435, opened 2019-11-21
Fix untranslated strings

naconner:


naconner:
[Translate ID scan text, 'More' & 'Less'](https://trello.com/c/f70SSEqb/105-translate-id-scan-text-more-less)

PR #436, opened 2019-11-22
rework: dispense batching

RafaelTaranto:


naconner:
[Incorrect logging on dispensed bills when batched](https://trello.com/c/TehUUPAr/104-incorrect-logging-on-dispensed-bills-when-batched)

lib/brain.js
```diff
@@ -3180,7 +3143,81 @@ Brain.prototype._physicalDispense = function _physicalDispense () {
   emit('billDispenserDispensing')
 
   const notesToDispense = optimizeDispense(notes, this.billDispenser.dispenseLimit)
-  dispenseGenerator(notesToDispense, this.tx, this.config.completedTimeout)
+  const batchAmount = notesToDispense.length
+
+  const batchDispense = ({ dispensed, current, error }) => {
```
joshmh: I think we should break out long functions like this to regular JavaScript functions outside of the function. Better to have byte-sized pieces of logic.

joshmh: Use more descriptive variable names: ``currentBatch``, ``dispenseRecords``.

lib/brain.js
```diff
@@ -3180,7 +3143,81 @@ Brain.prototype._physicalDispense = function _physicalDispense () {
...
+    // short circuit out if a error is found
+    if (error) {
+      return { dispensed, error }
+    }
+
+    const notes = notesToDispense[current - 1]
```
joshmh: Rename ``current - 1`` to descriptive variable. ``current - 1`` should be checked to ensure it's in range.

lib/brain.js
```diff
@@ -3180,7 +3143,81 @@ Brain.prototype._physicalDispense = function _physicalDispense () {
...
+
+    this.browser().send({ action: 'dispensing', dispenseBatch: { current, of: batchAmount } })
```
joshmh: Not for now, but we should consider replacing this kind of thing with an emitted event, like we do with the board manager. Same will be true of events destined for l-s. Of course, even better would be to have a petri net emit this.

lib/brain.js
```diff
@@ -3180,7 +3143,81 @@ Brain.prototype._physicalDispense = function _physicalDispense () {
...
+        this.browser().send({ action: screen, dispenseBatch: { current: current, of: batchAmount } })
+
+        return this.billDispenser.waitForBillsRemoved()
+          .then(() => response)
+          .catch(error => _.assign(response, { error }))
+      })
+      .catch(error => ({ dispensed, error }))
+  }
+
+  const batchesFinished = ({ dispensed, current, error }) => {
```
joshmh: Break out into own function and describe what it does.

lib/brain.js
```diff
@@ -3180,7 +3143,81 @@ Brain.prototype._physicalDispense = function _physicalDispense () {
...
+      if (tx.id !== _.get('id')(this.tx)) {
+        return
+      }
+      return this._completed(tx.id)
+    })
+  }
+
+  const tasks = [...Array(batchAmount).fill(batchDispense), batchesFinished]
+
+  pWaterfall(tasks, {
```
joshmh: Describe what this does.

PR #439, opened 2019-11-27
feat: leds dimmer and on for qrscan

RafaelTaranto:


lib/upboard/gaia/led-manager.js
```diff
@@ -23,7 +23,7 @@ function processBrain (event) {
       return pulseUp(LEDS.validatorLeds, lc.COLORS.red)
 
     case 'scanBayLightOn':
-      return solidUp(LEDS.scanBayLeds, lc.COLORS.white)
+      return solidUp(LEDS.scanBayLeds, lc.COLORS.grey)
```
joshmh: Let's called it "dimmed".

PR #449, opened 2019-12-04
show_log script

naconner:


naconner:
[New show_log script](https://trello.com/c/jMetAWYf/110-new-showlog-script)

PR #458, opened 2019-12-12
feat: nippon printer support

RafaelTaranto:


RafaelTaranto:
[Support receipt printing on Nippon](https://trello.com/c/b0lMrPz5/68-support-receipt-printing-on-nippon#!)

hardware/codebase/upboard/gaia/device_config.json
```diff
@@ -11,6 +11,10 @@
   "compliance": {
     "paperWallet": false
   },
+  "kioskPrinter": {
+    "maker": "Nippon",
```
joshmh: Maybe this should be ``model`` and be more specific, like ``nippon-xx-32-60mm``. What if we decide to use a different printer from Nippon in the future?

hardware/system/upboard/install_script.sh
```diff
@@ -14,6 +18,23 @@ if [ "$MACHINE" != "gaia" ] && [ "$MACHINE" != "sintra" ]; then
   exit 1
 fi
 
+if [ -z $3 ]; then
+  PRINTER='None'
+fi
+
+if [ "$PRINTER" != "None" ] && [ "$PRINTER" != "zebra" ] && [ "$PRINTER" != "zippon" ]; then
```
joshmh:
```suggestion
if [ "$PRINTER" != "None" ] && [ "$PRINTER" != "zebra" ] && [ "$PRINTER" != "nippon" ]; then
```

lib/printer/loader.js
```diff
@@ -0,0 +1,64 @@
...
+    })
+}
+
+function load () {
+  const printerCfg = deviceConfig.kioskPrinter || { address: '/dev/ttyJ4' }
+  const { maker, model } = printerCfg
+  const noPrinterError = new Error('No printer found')
+
+  if (deviceConfig.brain.mockPrinter) {
+    return Promise.resolve(require('../mocks/kioskPrinter'))
```
joshmh: Better to enclose all of this in a new Promise, otherwise it's too easy to forget to return Promise.resolve/reject somewhere.

```
return new Promise((resolve, reject) => { ... })
```

lib/printer/loader.js
```diff
@@ -0,0 +1,64 @@
...
+      // if not nippon printing will fail silently
+      console.log(err)
+      return nippon
+    })
+}
+
+function load () {
+  const printerCfg = deviceConfig.kioskPrinter || { address: '/dev/ttyJ4' }
+  const { maker, model } = printerCfg
+  const noPrinterError = new Error('No printer found')
```
joshmh: I think there should be two errors: NoPrinterError and NoPrinterConfiguredError. Also, I think new errors should be created at each usage to avoid reuse.

lib/printer/loader.js
```diff
@@ -0,0 +1,64 @@
...
+
+  if (deviceConfig.brain.mockPrinter) {
+    return Promise.resolve(require('../mocks/kioskPrinter'))
+  }
+
+  // No printer
+  if (maker === 'None') return Promise.reject(noPrinterError)
+
+  // with model  ===  old config version
+  // no maker    ===  gaia shipped without config
```
joshmh: Add more detailed info here on exactly what the issues were with the old shipments.

Something like this but with correct info:

If ``model`` exists, we still need to auto-detect because some shipments were sent out with Nippons but configured for Zebras or vice versa.

If ``maker`` doesn't exist, we still need to auto-detect because some shipments were sent out without a config.

PR #459, opened 2019-12-17
feat: nippon printer

RafaelTaranto:


RafaelTaranto:
[Support receipt printing on Nippon](https://trello.com/c/b0lMrPz5/68-support-receipt-printing-on-nippon#!)

PR #463, opened 2019-12-31
fix: change udev gaia path

RafaelTaranto:


naconner:
[fix: change udev Gaia path](https://trello.com/c/09dIoY8x/167-fix-change-udev-gaia-path)

PR #467, opened 2020-01-06
chore: add COP to bills

naconner:


naconner:
[chore: add COP to bills](https://trello.com/c/36CMcWiy/162-chore-add-cop-to-bills)

PR #468, opened 2020-01-06
chore: increase log archive size

naconner:
Clear all logs with `purge-logs` to set a baseline and reduce log archive size. (Helpful as `lamassu-browser` and others don't have timestamps, thus we should clear old entries.)

Increase log archive size for `show_log`, as we've run into longer archives that are corrupted upon upload.


naconner:
[chore: increase log archive size](https://trello.com/c/DLcR7JQS/164-chore-increase-log-archive-size)

PR #469, opened 2020-01-09
fix: log unknownCommands on ccnet

RafaelTaranto:


RafaelTaranto:
[CCNET Validator 'Disabled'](https://trello.com/c/UrwLsS8K/160-ccnet-validator-disabled)

PR #484, opened 2020-01-20
fix: kill all leds on machine startup

RafaelTaranto:


RafaelTaranto:
[Clear hung led processes](https://trello.com/c/T3VXp84V/157-clear-hung-led-processes)

PR #486, opened 2020-01-22
fix: small reliability safeguards on dispense flow

RafaelTaranto:


RafaelTaranto:
[Safeguards against dispense on 0 crypto_atoms](https://trello.com/c/a5ZkzsKu/172-safeguards-against-dispense-on-0-cryptoatoms)

PR #489, opened 2020-01-29
test: brain batch dispense

mautematico:
I think I finally understood how batchDispense is suposed to work and have writen a couple of (failling) tests for the BillDispense.factory mock.

This is a WIP so I will be pushing new commits and may be even rewriting my old ones

### DIFF
--- a/lib/printer/loader.js
+++ b/lib/printer/loader.js
@@ -3,8 +3,9 @@ const fs = require('fs')
 
 const deviceConfig = require('../../device_config.json')
 
-const zebra = require('./zebra')
-const nippon = require('./nippon')
+// not optimal, but anyway node will cache this: https://nodejs.org/api/modules.html#modules_require_cache
+const zebra = () => require('./zebra')
+const nippon = () => require('./nippon')
 
 const ZEBRA_MODEL = 'Zebra-KR-403'
 const NIPPON_MODEL = 'Nippon-2511D-2'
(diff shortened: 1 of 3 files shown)


mautematico:
It turns out that my test cases were wrong, I have now fixed and I'm adding new ones :)

mautematico:
[Write tests for the batch dispenser code](https://trello.com/c/s3gsPQVi/34-write-tests-for-the-batch-dispenser-code)

lib/printer/loader.js
```diff
@@ -28,18 +29,18 @@ function configureAsZebra (printerCfg) {
 
 function autoDetect (printerCfg) {
   // this can be either nippon, zebra or no printer
-  return zebra.checkStatus(printerCfg, 2000)
+  return zebra().checkStatus(printerCfg, 2000)
     .then(() => {
       // only zebras can respond
       configureAsZebra(printerCfg)
-      return zebra
+      return require('./zebra')
```
joshmh: Why not ``return zebra()``?

lib/printer/loader.js
```diff
@@ -63,11 +64,11 @@ function load () {
 
     // if no model: some gaias were shipped without a config
     if (!model) {
-      return resolve(zebra)
+      return resolve(require('./zebra'))
```
joshmh: Why not ``zebra()``?

lib/printer/loader.js
mautematico: oops. fixed

PR #491, opened 2020-02-04
fix: make supervisord be managed by systemctl

RafaelTaranto:


RafaelTaranto:
[Correct supervisord restart upon clock drifts](https://trello.com/c/1j8naFmM/27-correct-supervisord-restart-upon-clock-drifts)

PR #492, opened 2020-02-04
chore: make network error message friendly

RafaelTaranto:


RafaelTaranto:
[Change 'ephemeral network error' in machine logs to something less alarming](https://trello.com/c/K2ql24V2/5-change-ephemeral-network-error-in-machine-logs-to-something-less-alarming)

PR #493, opened 2020-02-04
fix: ignore invalid minimum tx value

RafaelTaranto:


RafaelTaranto:
[Don't fail when minimum transaction value is invalid](https://trello.com/c/VsIUhZZD/17-dont-fail-when-minimum-transaction-value-is-invalid)

PR #494, opened 2020-02-08
fix: correct printer capitalisation

naconner:


naconner:
[Correct printer capitalisation in install_script.sh](https://trello.com/c/VqUSVMup/40-correct-printer-capitalisation-in-installscriptsh)

PR #495, opened 2020-02-10
chore: update locales

naconner:


naconner:
[chore: update locales](https://trello.com/c/F524CZqk/41-chore-update-locales)

PR #496, opened 2020-02-12
fix: handle bill table response on ccnet protocol

RafaelTaranto:


RafaelTaranto:
[CCNET Validator 'Disabled'](https://trello.com/c/UrwLsS8K/42-ccnet-validator-disabled)

PR #498, opened 2020-02-17
fix: add billDispenser for upgraded Trofas

naconner:


naconner:
[Account for Santo Tirso on upboard/gaia device_config.json](https://trello.com/c/segYPcYc/3-account-for-santo-tirso-on-upboard-gaia-deviceconfigjson)

PR #501, opened 2020-02-26
fix: update acp chromium

RafaelTaranto:


RafaelTaranto:
[Improve Chromium screen maximisation on Aaeon ACP](https://trello.com/c/fArJ03pw/20-improve-chromium-screen-maximisation-on-aaeon-acp)

PR #504, opened 2020-03-15
fix: handle invalid chars on log files

RafaelTaranto:


RafaelTaranto:
[Find source of unexpected tokens in logs](https://trello.com/c/WQT3Dbnb/43-find-source-of-unexpected-tokens-in-logs)

PR #505, opened 2020-03-17
fix: facephoto as hardlimit when camera not found

RafaelTaranto:


RafaelTaranto:
[Do not bail out if customer camera KYC option enabled and machine doesn't have customer camera](https://trello.com/c/WVXE4dCQ/47-do-not-bail-out-if-customer-camera-kyc-option-enabled-and-machine-doesnt-have-customer-camera)

PR #509, opened 2020-03-20
fix: facephoto as hardlimit when camera not found

RafaelTaranto:


RafaelTaranto:
[Do not bail out if customer camera KYC option enabled and machine doesn't have customer camera](https://trello.com/c/WVXE4dCQ/47-do-not-bail-out-if-customer-camera-kyc-option-enabled-and-machine-doesnt-have-customer-camera)

lib/scanner.js
```diff
@@ -50,6 +51,19 @@ function isOpened () {
   return opened
 }
 
+function hasCamera (mode) {
+  const cameraConf = mode === 'facephoto' ? 'frontFacingCamera' : 'scanner'
+
+  const device = _.get(`${cameraConf}.device`, configuration)
+  const modeConfig = _.get([cameraConf, mode], configuration)
```
joshmh: Shouldn't we check for null device?

naconner:
[Account for Santo Tirso on upboard/gaia device_config.json](https://trello.com/c/segYPcYc/3-account-for-santo-tirso-on-upboard-gaia-deviceconfigjson)

lib/scanner.js
RafaelTaranto: Good point, the try catch would handle this case but I added a new test for it.

PR #512, opened 2020-03-26
feat: add udev rules on douros

RafaelTaranto:


RafaelTaranto:
[Support deployment of udev rules to aaeon-acp](https://trello.com/c/EaEokplo/89-support-deployment-of-udev-rules-to-aaeon-acp)

PR #513, opened 2020-03-31
fix: silently fail on leds kill all

RafaelTaranto:


RafaelTaranto:
[Clear hung led processes](https://trello.com/c/T3VXp84V/102-clear-hung-led-processes)

PR #522, opened 2020-07-15
fix: use correct username on browser supervisor

RafaelTaranto:


RafaelTaranto:
[Verify proper deployment with manual_update.js on UPboard](https://trello.com/c/qR33YMLy/154-verify-proper-deployment-with-manualupdatejs-on-upboard)

PR #523, opened 2020-07-20
feat: include model on pings

mautematico:
include model on pings, so it is saved for later display on machine status page


mautematico:
[Machine status fixes](https://trello.com/c/nERdgEIJ/144-machine-status-fixes)

PR #524, opened 2020-07-30
fix: hard limit with no ff camera on machine bug

RafaelTaranto:


RafaelTaranto:
[Fix customer camera settings regression](https://trello.com/c/OyQSAKTC/170-fix-customer-camera-settings-regression)

PR #525, opened 2020-08-05
fix: support multiple versions of chromium

RafaelTaranto:


RafaelTaranto:
[Add support back to older versions of chromium on aaeon ACP](https://trello.com/c/1etKAymW/175-add-support-back-to-older-versions-of-chromium-on-aaeon-acp)

PR #526, opened 2020-08-09
fix: use correct name for camera test

RafaelTaranto:


RafaelTaranto:
[douro1 still requesting user photo on v7.4.9](https://trello.com/c/FhPceyNi/178-douro1-still-requesting-user-photo-on-v749)

PR #532, opened 2020-09-18
fix: preserve cryptomatModel upon update

naconner:
Fixes issue with updating the Douro II, where `cryptomatModel` was overwritten, as lamassu-watchdog only conveys the platform on the ssuboard, rather than platform + model as it does with UPboards.

For UPboards, a further safeguard against overwriting.


naconner:
[Preserve cryptomatModel upon update](https://trello.com/c/OySL4h0k/221-preserve-cryptomatmodel-upon-update)

PR #533, opened 2020-09-25
feat: warn when front facing camera is not found

RafaelTaranto:


RafaelTaranto:
[Don't bail out l-m if facephoto camera is undefined](https://trello.com/c/1lCEq27k/220-dont-bail-out-l-m-if-facephoto-camera-is-undefined)

PR #541, opened 2020-10-23
feat: relay machine model

RafaelTaranto:


RafaelTaranto:
[Pass down machine model to server](https://trello.com/c/ifRaZFKT/280-pass-down-machine-model-to-server)

PR #550, opened 2020-11-03
fix: 'Shutdown' admin button not shutting down machine

liordino:
feat: created shutdown action

### DIFF
--- a/lib/brain.js
+++ b/lib/brain.js
@@ -425,6 +425,7 @@ Brain.prototype._initTraderEvents = function _initTraderEvents () {
   this.trader.on('error', function (err) { console.log(err.stack) })
   this.trader.on('unpair', function () { self._unpair() })
   this.trader.on('reboot', function () { self._reboot() })
+  this.trader.on('shutdown', function () { self._shutdown() })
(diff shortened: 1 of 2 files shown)


liordino:
['Shutdown' admin button not shutting down machine](https://trello.com/c/4GdLCkw0/320-shutdown-admin-button-not-shutting-down-machine)

RafaelTaranto (approved):

lib/brain.js
```diff
@@ -2945,6 +2946,13 @@ Brain.prototype._reboot = function _reboot () {
   })
 }
 
+Brain.prototype._shutdown = function _shutdown () {
+  console.log('Remote shutdown')
+  cp.execFile('shutdown', ['now'], {}, function () {
+    process.exit(0)
```
joshmh, line 2952: Shouldn't we attempt a more graceful shutdown, like only when the machine is idle?

joshmh (changes requested):

lib/brain.js
joshmh: Decided to do this is an separate card.

PR #554, opened 2020-11-10
fix: compliance for backwards compatible server

RafaelTaranto:


RafaelTaranto:
[Compliance flow on l-m 7.5 while on l-s 7.4](https://trello.com/c/3hvS58lb/357-compliance-flow-on-l-m-75-while-on-l-s-74)

PR #555, opened 2020-11-17
feat: validate bech32 zec addresses

naconner:
Depends on https://github.com/lamassu/lamassu-server/pull/518

### DIFF
--- a/lib/coins/zec.js
+++ b/lib/coins/zec.js
@@ -40,4 +40,5 @@ function validate (network, address) {
   if (!address) throw new Error('No address supplied.')
   if (base58Validator(network, address, base58Opts)) return true
   if (zecBech32Validator(network, address, bech32Opts)) return true
+  return false
 }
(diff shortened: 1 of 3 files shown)


RafaelTaranto (approved):

lib/coins/zec.js
```diff
@@ -32,5 +38,6 @@ function depositUrl (address, amount) {
 function validate (network, address) {
   if (!network) throw new Error('No network supplied.')
   if (!address) throw new Error('No address supplied.')
-  return base58Validator(network, address, base58Opts)
+  if (base58Validator(network, address, base58Opts)) return true
+  if (zecBech32Validator(network, address, bech32Opts)) return true
```
joshmh, line 42: Need to return ``false`` if the other tests fail. Also do we have a test for failing this validation function? It should have caught that.

joshmh (changes requested):

RafaelTaranto (approved):

PR #559, opened 2020-11-24
Feat: Machine will wait for a transaction to finish before shutting down

csrapr:


csrapr:
[Shutdown by admin action should be graceful](https://trello.com/c/NuBA46ut/339-shutdown-by-admin-action-should-be-graceful)

lib/brain.js
```diff
@@ -2956,6 +2956,11 @@ Brain.prototype._reboot = function _reboot () {
 }
 
 Brain.prototype._shutdown = function _shutdown () {
+  // can only run if machine is no in the middle of a transaction
+  if (!this.isStaticState()) {
+    console.log ("In the middle of a transaction. Will shutdown when it is done...")
```
RafaelTaranto, line 2961: How will this shutdown when done? Looks like you are just ignoring the request.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #566, opened 2021-01-08
fix: cash-out receipt printed even if disabled

RafaelTaranto:


RafaelTaranto:
[Cash-out receipt being printed regardless of config](https://trello.com/c/m6ZEZ088/507-cash-out-receipt-being-printed-regardless-of-config)

jsousa127 (approved):

PR #567, opened 2021-01-14
Fix wallet address format case sensitive

chaotixkilla:


chaotixkilla:
[Wallet address format case sensitive (bech32)](https://trello.com/c/Q7XXgh4y/475-wallet-address-format-case-sensitive-bech32)

lib/coins/btc.js
```diff
@@ -17,7 +17,7 @@ const bech32Opts = {
 module.exports = {depositUrl, parseUrl, formatAddress, base58Opts, bech32Opts, createWallet}
 
 function parseUrl (network, url) {
-  const res = /^([bB]itcoin:\/{0,2})?(\w+)/.exec(url)
+  const res = /^(bitcoin:\/{0,2})?(\w+)/i.exec(url)
```
csrapr, line 20: check if it's necessary to allow between 0 and 2 forward slashes between the first capture group and the second

csrapr (changes requested):

csrapr (approved):

RafaelTaranto (changes requested):
Why are the app.js files being changed here? lib/coins/btc.js should only be used by the nodejs part of the code. remove the builded changes.

RafaelTaranto (approved):

PR #568, opened 2021-01-15
Pass raw ID data

chaotixkilla:


chaotixkilla:
[Passing along the raw data for IDs](https://trello.com/c/68Vbflqn/513-passing-along-the-raw-data-for-ids)

RafaelTaranto (approved):

PR #572, opened 2021-01-27
Fix cards without country attribute defaulting to 'USA'

chaotixkilla:


chaotixkilla:
[ID Data parses Quebec licenses as 'USA'](https://trello.com/c/gT4IAFlI/537-id-data-parses-quebec-licenses-as-usa)

lib/compliance/parsepdf417-us.js
```diff
@@ -20,6 +20,15 @@ var CODES = {
...
 
+const states = {
+  'CA': ['AB', 'BC', 'MB', 'NB', 'NL', 'NT', 'NS', 'NU', 'ON', 'PE', 'QC', 'SK', 'YT'],
+  'USA': ['AL', 'AK', 'AZ', 'AR', 'CA', 'CO', 'CT', 'DE', 'FL', 'GA',
+    'HI', 'ID', 'IL', 'IN', 'IA', 'KS', 'KY', 'LA', 'ME', 'MD',
+    'MA', 'MI', 'MN', 'MS', 'MO', 'MT', 'NE', 'NV', 'NH', 'NJ',
+    'NM', 'NY', 'NC', 'ND', 'OH', 'OK', 'OR', 'PA', 'RI', 'SC',
+    'SD', 'TN', 'TX', 'UT', 'VT', 'VA', 'WA', 'WV', 'WI', 'WY']
+}
+
```
csrapr, line 31: Two letter abbreviations for U.S Territories are missing, and also the letter abbreviation for District of Columbia (DC)
We should ask if these abbreviations should also be included or not
https://www.ncbi.nlm.nih.gov/books/NBK7254/

lib/compliance/parsepdf417-us.js
```diff
@@ -63,6 +72,12 @@ function parseRaw (data) {
   return normalize(result, version, issuingAuthorityNumber)
 }
 
+function getCountry (state) {
+  return _.findKey(s => {
+    return _.includes(state, s)
+  }, states)
+}
+
```
csrapr, line 80:
```suggestion
const getCountry = (state) => _.findKey(_.includes(state), states)
```

csrapr (changes requested):

lib/compliance/parsepdf417-us.js
chaotixkilla: I added them, just in case. Shouldn't be much of a problem either way.

lib/compliance/parsepdf417-us.js
chaotixkilla: Yeah, that would make more sense, thanks!

csrapr (approved):

lib/compliance/parsepdf417-us.js
```diff
@@ -79,7 +93,7 @@ function normalize (result, version, issuingAuthorityNumber) {
   if (result.firstName) result.firstName = result.firstName.split(',')[0]
   var date = result.dateOfBirth.replace(/-/g, '')
   var expirationDate = result.expirationDate ? result.expirationDate.replace(/-/g, '') : null
-  result.country = result.country || 'USA'
+  result.country = result.country || getCountry(result.state)
```
RafaelTaranto, line 96: Might be better to still leave 'USA' as the default. There might be some problem states out of spec.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #573, opened 2021-01-28
Parse BCID cards

chaotixkilla:


chaotixkilla:
[ID Data does not parse British Columbia bar codes](https://trello.com/c/PLlgeQ9j/536-id-data-does-not-parse-british-columbia-bar-codes)

csrapr (approved):

csrapr (changes requested):
The person in the example code weighs 95 pounds, but what if they weighed 195 pounds?
I tested the code and the output is this: `weight: "19"` (notice how it is ignoring the trailing 5)
Also if the weight is 195 the eye color becomes "NBL"

try testing here:
https://npm.runkit.com/lodash
https://paste.ofcode.org/32KtvRjtPtrJhtCFdkBGA3f

csrapr (approved):

lib/compliance/parsepdf417.js
```diff
@@ -5,6 +5,7 @@ const crypto = require('crypto')
 const _ = require('lodash/fp')
 
 const parsers = [
+  require('./parsepdf417-ca-bc'),
```
RafaelTaranto, line 8: isn't it better to check for us first?

RafaelTaranto (changes requested):

RafaelTaranto (approved):

lib/compliance/parsepdf417-ca-bc.js
```diff
@@ -0,0 +1,94 @@
...
+  const track2Start = data.indexOf(';')
+  const track3Start = data.indexOf('_%')
+
+  if (track1Start === -1 || !track2Start === -1 || !track3Start === -1) return null
+
+  return parseRaw(data)
+}
+
+function parseRaw (data) {
+  const track1Start = data.indexOf('%')
```
joshmh, line 14: We shouldn't do these searches twice. If we merge parse and parseRaw, the combined function is just one line longer than parseRaw (the ``if`` statement).

joshmh (changes requested):

PR #577, opened 2021-02-05
Verify customer suspend state when tx

josepfo:


josepfo:
["Suspend" state not showing on admin](https://trello.com/c/vee9n3T2/527-suspend-state-not-showing-on-admin)

chaotixkilla (approved):
Seems good, really like the JSDoc comments. Perhaps it is something we should be doing more

lib/brain.js
```diff
@@ -1546,6 +1546,96 @@ Brain.prototype.twoWayMode = function twoWayMode () {
...
+ * one of the enabled compliance tier thresholds
+ *
+ * @name isSuspended
+ * @function
+ *
+ * @param {object} customer Acting customer
+ * @param {date} now Current date
+ * @returns {bool} Whether customer is suspended or not
+ */
+Brain.prototype.isSuspended = function isSuspended (customer, now) {
```
RafaelTaranto, line 1562: I don't see a reason why now is a param, can't we just create a new Date within this function?

lib/brain.js
```diff
@@ -1546,6 +1546,96 @@ Brain.prototype.twoWayMode = function twoWayMode () {
...
+ * else the bill screen will be displayed with the relative error message
+ *
+ * @name showSuspendedCustomer
+ * @function
+ *
+ * @param {object} customer Acting customer
+ * @param {date} now Current date
+ *
+ */
+Brain.prototype.showSuspendedCustomer = function showSuspendedCustomer (customer, now) {
```
RafaelTaranto, line 1579: Same as above

RafaelTaranto (approved):

lib/brain.js
```diff
@@ -1546,6 +1546,96 @@ Brain.prototype.twoWayMode = function twoWayMode () {
...
+* current date and the suspension time
+*
+* @name hardLimitReached
+* @function
+*
+* @param {object} customer Acting customer
+* @param {date} now Current date
+* @returns {object} hard limit
+*/
+Brain.prototype.hardLimitReached = function hardLimitReached (customer, now) {
```
RafaelTaranto, line 1623: Can we rename this? By the name looks like we're testing for a boolean. 

Maybe `getHardLimitReachedData`

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #580, opened 2021-02-10
Lowercasing bech32 addresses

chaotixkilla:


chaotixkilla:
[Set bech32 all-caps addresses to lowercase](https://trello.com/c/1OEKXHTX/569-set-bech32-all-caps-addresses-to-lowercase)

josepfo (approved):

RafaelTaranto (approved):

lib/coins/btc.js
```diff
@@ -24,7 +25,9 @@ function parseUrl (network, url) {
 
   if (!validate(network, address)) throw new Error('Invalid address')
 
-  return address
+  return isBech32Address(address, bech32Opts)
```
joshmh, line 28: This should be generalized more. We need a new ``formatAddress`` function that takes an address and lowercases it when necessary.

joshmh (changes requested):

PR #586, opened 2021-02-21
fix: add escape characters to i18n quotes

naconner:


chaotixkilla (approved):

i18n/ui/lbm-ui_en-US.po
```diff
@@ -475,8 +475,8 @@ msgid "Please press the back of your driver's license against the scan window."
 msgstr "Please press the back of your driver's license against the scan window."
 
 #: On screen: scan-photo-manual
-msgid "Please press the front of your state ID card against the scan window, hold it there, and then press "Ready to scan"."
-msgstr "Please press the front of your state ID card against the scan window, hold it there, and then press "Ready to scan"."
+msgid "Please press the front of your state ID card against the scan window, hold it there, and then press \"Ready to scan\"."
+msgstr "Please press the front of your state ID card against the scan window, hold it there, and then press \"Ready to scan\"."
```
chaotixkilla, line 479: The character escaping should be done on ui/html/scan-photo-manual.html, otherwise it will revert to non-escaped characters once compiling happens
Once that is done, run "npm run watch" and everything should be compiled and working

chaotixkilla (changes requested):

i18n/ui/lbm-ui_en-US.po
naconner: Is `lbm-ui_en-US.po` generated from `scan-photo-manual.html`? I cannot upload this file to Crowdin without these being escaped here, as it rejects it otherwise.

chaotixkilla: The strings on `lbm-ui_en-US.po` are generated by x-raying the HTML files, and the source for that specific string is the `scan-photo-manual.html` file. Sorry for the delayed response

RafaelTaranto (approved):

PR #587, opened 2021-03-01
Relay machine name in lamassu-update

josepfo:


josepfo:
[Relay machine name in lamassu-update](https://trello.com/c/x7ZL68cm/608-relay-machine-name-in-lamassu-update)

lib/update/updater.js
```diff
@@ -181,9 +181,23 @@ Updater.prototype._update = function _update () {
...
 
+Updater.prototype.updateHeaders = function updateHeaders (options) {
+  var dataPath = path.resolve(__dirname, '..', this.config.dataPath)
+  var machineInfo = this.machineInfo.load(dataPath)
+
+  if (machineInfo.active) {
+    return _.assign(options, {
+      'device-machine-id': machineInfo.deviceId,
+      'device-machine-name': machineInfo.deviceName
+    })
```
csrapr, line 192: what's the reason behind having -machine- in the mddle? We usually call machines "devices" throughout our codebase

csrapr (changes requested):

lib/update/updater.js
josepfo: httpsOptions.headers already have a property 'device-id', the -machine- is pretty much to differentiate

RafaelTaranto (approved):

lib/trader.js
```diff
@@ -407,6 +409,8 @@ Trader.prototype.pollHandler = function pollHandler (res) {
     this.triggers = createCompatTriggers(res)
   }
 
+  machineInfo.save(this.dataPath, res.machineInfo)
```
joshmh, line 412: Shouldn't we be using ``await`` here?

joshmh (changes requested):

lib/trader.js
RafaelTaranto: Machine info isn't required for the machine to work, so we can move past this without waiting. Also this is inside the polling function, waiting for FS work should be prevented.

RafaelTaranto (approved):

PR #588, opened 2021-03-02
Ignore timeout when waiting for bills

josepfo:


josepfo:

lib/f56/f56-dispenser.js
```diff
@@ -61,6 +61,7 @@ BillDispenser.prototype.billsPresent = function billsPresent () {
 
 BillDispenser.prototype.waitForBillsRemoved = function waitForBillsRemoved () {
   return new Promise((resolve, reject) => {
+    var retries = 3
```
csrapr, line 64: I think using var is disallowed by the linter, but this is lamassu-machine so, not sure.

```suggestion
    let retries = 3
```

csrapr (changes requested):

lib/f56/f56-dispenser.js
josepfo: Hmm, I just checked and there's let/var in multiple files. Do you think it's worth changing?

csrapr: nah

RafaelTaranto (changes requested):

lib/f56/f56-rs232.js
```diff
@@ -134,6 +134,11 @@ function request (command) {
       rs232StatusPointer.off()
       statusPointer.off()
       if (status === 'Response') return resolve(frame)
+      if (status === 'Response Timeout') {
+        const error = new Error('Can\'t send in state: ' + dLevelFsm.state)
```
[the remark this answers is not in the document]
josepfo: If the amount reaches 4 digits it goes to a new line. Should I increase the width even more?

RafaelTaranto (approved):

PR #593, opened 2021-03-11
Fix: fix QR code not reading

csrapr:


csrapr:
[Bitpay QR code not scanning](https://trello.com/c/2Iya7581/511-bitpay-qr-code-not-scanning)

package.json
```diff
@@ -3,6 +3,7 @@
   "license": "Unlicense",
   "version": "7.5.0-beta.2",
   "dependencies": {
+    "@joepie91/v4l2camera": "^1.0.5",
```
chaotixkilla, line 6: On line `49`, there's the same dependency, but on another version and possibly repo. Check with Rafael whether that optionalDependency should be removed or not and if it is worth for us to create our own fork of the v412camera repository

chaotixkilla (changes requested):

chaotixkilla (approved):

RafaelTaranto (approved):

PR #595, opened 2021-03-12
feat: validate bech32 zec

josepfo:


josepfo:
[Shielded ZEC support](https://trello.com/c/9effiH90/232-shielded-zec-support)

lib/coins/validators/index.js
```diff
@@ -61,6 +62,28 @@ function bech32Validator (network, address, opts) {
...
+  const data = bech32.fromWords(decoded.words)	
+  if (data.length !== 43) {	
+    console.log(`Invalid bech32 address length: ${data.length}`)	
+    return false	
+  }
+
+  if (network === 'main' && decoded.prefix === opts.mainNetPrefix) return true
+  if (network === 'test' && decoded.prefix === opts.testNetPrefix) return true
+  return false
+}
```
chaotixkilla, line 88: This function has a lot of code equal to the `bech32Validator()` function. Consider merging this function with `bech32Validator()` and keep it all under the `bech32Validator` name

chaotixkilla (changes requested):

lib/coins/validators/index.js
RafaelTaranto: I agree that there's some repeated code here, but it's doing a different thing in the end. If the witness version test and lenght are different a function with a "flag" would be harder to read.

chaotixkilla (approved):

PR #597, opened 2021-03-16
fix: properly load machine info

RafaelTaranto:


RafaelTaranto:
[Relay machine name in lamassu-update](https://trello.com/c/x7ZL68cm/608-relay-machine-name-in-lamassu-update)

PR #601, opened 2021-03-17
feat: capture multiple id card photos

josepfo:


josepfo:
[Capture and relay ID Data stills to operator server if not parsed](https://trello.com/c/U7bZ7bNI/607-capture-and-relay-id-data-stills-to-operator-server-if-not-parsed)

lib/scanner.js
```diff
@@ -154,12 +154,33 @@ function scanQR (callback) {
   })
 }
 
-function scanPDF417 (callback) {
+function scanPDF417 (callback, idCardStillsCallback) {
+  let captureIdCardPhoto = false
+  let idCardTimer = 4000
+  let endInterval = 24000
```
chaotixkilla, line 160: These timers could be moved to `software_config.json`. Probably is not a huge deal, it's your choice.

chaotixkilla: I don't see `endInterval` being reassigned anywhere, so it could be changed to `const`

lib/scanner.js
```diff
@@ -154,12 +154,33 @@ function scanQR (callback) {
...
+      captureIdCardPhoto = false
+      return _callback(null, null)
+    }
     var parsed = Pdf417Parser.parse(result)
-    if (!parsed) return _callback(null, null)
+    if (!parsed) {
+      if (captureIdCardPhoto) idCardStillsCallback(frame)
+      captureIdCardPhoto = false
+      return _callback(null, null)
+    }
```
chaotixkilla, line 182: A bit nitpicking, but 

```
if (captureIdCardPhoto) idCardStillsCallback(frame)
captureIdCardPhoto = false
```

could be moved to a `takePhoto()` function

lib/compliance/flows/id-card-data.js
```diff
@@ -10,34 +10,44 @@ const idDataMachine = xstate.Machine({
...
 })
 
 let currentStateValue
 let data
+let idCardDataStills = []
 
 function getData () { return data }
 function setData (value) { data = value }
 
+function setIdDataPhotos (photo) {
```
chaotixkilla, line 42: Hmm, function name should be `addIdDataPhoto()`, as we're not really setting anything, we're adding to an array

lib/scanner.js
```diff
@@ -154,12 +154,33 @@ function scanQR (callback) {
...
 
-function scanPDF417 (callback) {
+function scanPDF417 (callback, idCardStillsCallback) {
+  let captureIdCardPhoto = false
+  let idCardTimer = 4000
+  let endInterval = 24000
+  let handleIdCardPhoto = setInterval(() => {
+    captureIdCardPhoto = true
+    idCardTimer = idCardTimer + 4000
+  }, idCardTimer)
```
chaotixkilla, line 164: Incrementing the timer like this would make the interval increasingly longer. Your `idCardTimer` has the following values on each iteration [4000, 8000, 12000, 16000, 20000]. This would work if you were working with `setTimeout()`, but on intervals, you only need to pass the timer for each iteration, so no need to increment it.
I would do something slightly different, something of this sort:

```
const ID_CARD_CAPTURE_INCREMENT = 4000
const PHOTO_LIMIT = 5
let photosTaken = 0
const handleIdCardPhoto = setInterval(() => {
  captureIdCardPhoto = true
  photosTaken++
}, ID_CARD_CAPTURE_INCREMENT)

...

if (photosTaken >= PHOTO_LIMIT) clearInterval(handleIdCardPhoto)
```

chaotixkilla (changes requested):

lib/scanner.js
josepfo: Actually, I think the interval stays the same because it uses the initial timer value when setInterval() is called. But it's way cleaner what you suggested. Thanks!

lib/compliance/flows/id-card-data.js
```diff
@@ -10,34 +10,42 @@ const idDataMachine = xstate.Machine({
...
 let currentStateValue
 let data
+let idCardDataStills = []
 
 function getData () { return data }
 function setData (value) { data = value }
 
+function addIdDataPhoto (photo) {
+  idCardDataStills.push(photo)
+  if (_.size(idCardDataStills) > 5) idCardDataStills = idCardDataStills.slice(5)
```
chaotixkilla, line 44: If you want to take the last 5 members of `idCardDataStills`, I think you can just do

```suggestion
  if (_.size(idCardDataStills) > 5) idCardDataStills = _.takeRight(5, idCardDataStills)
```

chaotixkilla (changes requested):

lib/brain.js
```diff
@@ -660,12 +660,26 @@ Brain.prototype.timeoutToScannerCancel = function timeoutToScannerCancel (stateM
...
     this.startDisabled = false
 
+    if (err || !result) {
+      const photos = idCardData.getIdDataPhotos()
+      this.trader.updateIdCardPhotos(customer.id, { photos })
+        .then(() => console.log('Successfully saved ', photos.length, ' id data photos'))
+        .catch(err => {
+          console.log('Error saving id card photos', err)
+        })
+    }
```
RafaelTaranto, line 681: Don't we have that problem here were we pass through this code multiple times during a single id scan try?

lib/compliance/flows/id-card-data.js
RafaelTaranto, line 44: I would prefer to have brackets and new lines here if we're doing a assignment. I also think this can be simplified

```js
  if (_.size(idCardDataStills) <= 5) {
    idCardDataStills.push(photo)
  }
```

RafaelTaranto (changes requested):

lib/compliance/flows/id-card-data.js
josepfo: It is but we wouldn't prune the array and populate it with new photos upon multiple scan retries

RafaelTaranto (approved):

PR #603, opened 2021-03-18
Fix increase timeout on f56

josepfo:


josepfo:

lib/f56/f56-dlevel-fsm.js
```diff
@@ -2,7 +2,7 @@
 
 const machina = require('machina')
 
-const RESPONSE_TIMEOUT = 20000
+const RESPONSE_TIMEOUT = 30000
```
joshmh, line 5:
```suggestion
const RESPONSE_TIMEOUT = 40000
```

Let's make it 40 seconds.

joshmh (changes requested):

PR #613, opened 2021-04-29
fix: escape double quotes and keep html tags

josepfo:


josepfo:
[Fix translation file generation](https://trello.com/c/jp8flnlV/718-fix-translation-file-generation)

chaotixkilla (approved):

i18n/ui/lbm-ui_en-US.po
```diff
@@ -346,6 +346,10 @@ msgstr "We're all out of coins!"
 msgid "Please try again later."
 msgstr "Please try again later."
 
+#: On screen: limit-reached
+msgid "<div> OK </div>"
+msgstr "<div> OK </div>"
+
```
RafaelTaranto, line 352: Hehe, you just found a bug on the limit-reached screen. i18n tag should be on the inner div.

i18n/ui/lbm-ui_en-US.po
```diff
@@ -527,8 +531,8 @@ msgid "Initializing hardware."
 msgstr "Initializing hardware."
 
 #: On screen: booting
-msgid "We'll be up in a moment."
-msgstr "We'll be up in a moment."
+msgid "<strong>We'll be up in a moment.</strong>"
```
RafaelTaranto, line 534: Having the tags on the ID should change what's translatable based on the locales.js file. Can you verify that this string gets translated to another language?

RafaelTaranto (changes requested):

PR #615, opened 2021-05-18
Feat/arbitrary compliance NUMERICAL

csrapr:


csrapr:
[Custom requirements - numerical](https://trello.com/c/8mHSmuqW/746-custom-requirements-numerical)

csrapr:
[Custom requirements - numerical](https://trello.com/c/8mHSmuqW/746-custom-requirements-numerical)

lib/brain.js
```diff
@@ -11,7 +11,7 @@ const pify = require('pify')
 const pDelay = require('delay')
 const pAny = require('p-any')
 const pSettle = require('p-settle')
-
+const uuid = require('uuid');
```
chaotixkilla, line 14: Extra `;` at the end of this line

lib/brain.js
```diff
@@ -638,11 +645,28 @@ Brain.prototype.transitionScreen = function transitionScreen () {
           break
       }
     }
+
+    if (!appState) {
+      // custom info request state
+      machineState = customTier.getState()
+      switch (machineState) {
+        case 'customInfoRequestScreen2':
+          appState = 'customInfoRequestScreen2'
```
chaotixkilla, line 654: Could this be changed to a more expressive screen name? Instead of enumerating the possible custom info request screens as '1', '2', '3', etc., having something like 'frontPage', 'numericInput', etc.

lib/brain.js
```diff
@@ -1883,6 +1948,14 @@ Brain.prototype.isTierCompliant = function isTierCompliant (tier) {
...
     case 'suspend':
       return false
+    case 'custom':
+      const customInfoRequestId = tier
+      const customerData = _.find(['info_request_id', customInfoRequestId])(customer.customInfoRequestData)
+      if (!customerData) return false
+      if (customerData.approved === true) return true
+      if (customerData.approved === false) return false
+      // when approved is null then data was already added but not manually approved yet by the operator
+      return true
```
chaotixkilla, line 1958: Could this be replaced with `return !!customerData?.approved || !!customerData`?

ui/js/keypad.js
```diff
@@ -170,4 +196,24 @@ Keypad.prototype._keyPress = function _keyPress (target) {
...
+      case 'length':
+        LENGTHS.custom = this.opts.maxLength
+        break
+      case 'date':
+        LENGTHS.custom = 8
+        break
+      default:
+        break
+    }
+  }
```
chaotixkilla, line 218: Wouldn't it be better to just ditch this function when you find out `this.opts.constraint` is not set?

```suggestion
Keypad.prototype.setOpts = function setOpts (newOpts) {
  this.opts = newOpts
  if (!this.opts.constraint) return
  
  switch (this.opts.constraint) {
      case 'length':
        LENGTHS.custom = this.opts.maxLength
        break
      case 'date':
        LENGTHS.custom = 8
        break
      default:
        break
    }
```

ui/src/app.js
```diff
@@ -279,6 +280,12 @@ function processData (data) {
     case 'invalidPromoCode':
       setState('promo_code_not_found')
       break
+    case 'customInfoRequest':
+      customInfoRequest(data.customInfoRequest, 1)
+      break
+    case 'customInfoRequestScreen2':
+      customInfoRequest(data.customInfoRequest, 2)
+      break
```
chaotixkilla, line 288: This is related to the comment of having more explicit screen names

ui/src/app.js
```diff
@@ -292,6 +299,26 @@ function usSsnPermission () {
...
+    customRequirementNumericalKeypad.setOpts({
+      type: 'custom',
+      constraint: customInfoRequest.input.constraintType,
+      maxLength: customInfoRequest.input.numDigits
+    })
+    customRequirementNumericalKeypad.activate()
+    setState('custom_permission_screen2')
+    setScreen('custom_permission_screen2')
+  }
+}
```
chaotixkilla, line 320: Same here

chaotixkilla (changes requested):
Some small changes, shouldn't take long

lib/brain.js
csrapr: I'll see if I can make the screen 2 more explicit for each type of input, but the reason is because that's how the wizard organizes info into screen 1 and screen 2.
on the db we also have stuff saved like that:

````JSON
{
   "name":"Birth date",
   "input":{
      "type":"numerical",
      "constraintType":"date"
   },
   "screen1":{
      "text":"On the next screen please enter your birth date.",
      "title":"We need to know your birth date"
   },
   "screen2":{
      "text":"Enter your birth date",
      "title":"Birth date"
   }
}
````

File: lib/brain.js [hunk 7]
```diff
@@ -1951,9 +1950,7 @@ Brain.prototype.isTierCompliant = function isTierCompliant (tier) {
     case 'custom':
       const customInfoRequestId = tier
       const customerData = _.find(['info_request_id', customInfoRequestId])(customer.customInfoRequestData)
-      if (!customerData) return false
-      if (customerData.approved === true) return true
-      if (customerData.approved === false) return false
+      if (!customerData || customerData.approved === false) return false
```
chaotixkilla, line 1953: If `customerData` always contains the field `approved`, this can just be replaced to `!!customerData.approved` I think

chaotixkilla (changes requested):

chaotixkilla (approved):

csrapr:
lots of rebases because some small change always escaped my undos, but it's only rebasing commits after the latest review.

PR #619, opened 2021-05-28
Fix BCID expiration date

chaotixkilla:


chaotixkilla:
[ID Data does not parse British Columbia bar codes](https://trello.com/c/PLlgeQ9j/536-id-data-does-not-parse-british-columbia-bar-codes)

PR #620, opened 2021-05-28
Fix machine bailing when expirationDate field is empty

chaotixkilla:


chaotixkilla:

PR #623, opened 2021-05-30
Fix machine bailing when expirationDate field is empty

chaotixkilla:


chaotixkilla:

RafaelTaranto (approved):

lib/brain.js
```diff
@@ -690,6 +690,9 @@ Brain.prototype.hasExpired = function hasExpired (cardData) {
   // TODO: ZA cards currently do not have an expiration date to confirm against
   if (cardData.country === 'ZA') return false
 
+  // In case the expiration date field is not found. Prevents l-m from bailing
+  if (_.isNil(cardData.expirationDate)) return false
```
joshmh, line 694: Can't a user just cover up their expiration date to make this succeed? Maybe in this case it should return ``true``.

joshmh (changes requested):

lib/brain.js
chaotixkilla: You're correct, seems obvious the edge case should be treated that way, now that I think about it

PR #625, opened 2021-06-08
fix: clear tx-db upon unpairing

josepfo:


josepfo:
[Upon unpairing, clear machine's existing tx-db](https://trello.com/c/7zt2HCOf/808-upon-unpairing-clear-machines-existing-tx-db)

PR #626, opened 2021-06-08
Fix: upon unpairing clear machine tx db

josepfo:


lib/db.js
```diff
@@ -93,3 +94,12 @@ function load (dbPath) {
       return _.sortBy(tx => tx.deviceTime, _.values(txTable))
     })
 }
+
+function clean (dbRoot) {
+  const files = fs.readdirSync(dbRoot)
+  if (files.length) {
+    cp.exec('rm ' + dbRoot + '/*', {}, err => {
```
RafaelTaranto, line 101: Is dbRoot guaranteed to never be null? This feels a bit risky

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #629, opened 2021-06-09
Fix: translation file generation 

josepfo:


josepfo:
[Fix translation file generation](https://trello.com/c/jp8flnlV/718-fix-translation-file-generation)

ui/html/limit-reached.html
```diff
@@ -16,7 +16,7 @@ <h3 class="js-i18n sm-margin-bottom">We're all out of coins!</h3>
     <button
       id="limit-reached-ok"
       display="inline-block"
-      class="filled-progress-button button ok js-i18n"
+      class="filled-progress-button button ok"
```
RafaelTaranto, line 19: The js-i18n should be on the inner div, here we're not translating this button at all.

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #631, opened 2021-06-18
PDF417 gender parsing

chaotixkilla:


chaotixkilla:
[Gender not shown on Customer Panel](https://trello.com/c/StvtZQfH/837-gender-not-shown-on-customer-panel)

lib/compliance/parsepdf417-us.js
```diff
@@ -101,6 +107,8 @@ function normalize (result, version, issuingAuthorityNumber) {
     result.expirationDate = expirationDate ? expirationDate.substr(4, 4) + expirationDate.substr(0, 4) : expirationDate
   }
 
+  result.gender = genders[result.gender]
+
```
RafaelTaranto, line 111: Looks like the canadian version is using 'sex' instead of 'gender'. Change parsepdf417-ca-bc to stay consistent. Also don't we have gendder for parsepdf417-za?

RafaelTaranto (changes requested):

PR #633, opened 2021-06-25
Add date format exceptions on pdf417 parsing

chaotixkilla:


chaotixkilla:

PR #642, opened 2021-07-19
Fix: cash out buttons active when limit is reached

josepfo:


josepfo:
[Cash-out allows selecting more than is allowed](https://trello.com/c/r18EqgNY/884-cash-out-allows-selecting-more-than-is-allowed)

chaotixkilla (approved):

RafaelTaranto (approved):

PR #643, opened 2021-07-20
feat: register if the manual trigger data is gathered

josepfo:


josepfo:
[Implement the manual option for triggers PT1](https://trello.com/c/2GUd8oIC/840-implement-the-manual-option-for-triggers-pt1)

lib/brain.js
```diff
@@ -682,6 +686,9 @@ Brain.prototype.scanPDF = function scanPDF () {
     }
 
     idCardData.setData(result)
+    if (_.get('idCardData')(this.trader.triggersAutomation) === MANUAL) {
+      this.manualTriggersDataProvided.idCardData = true
+    }
```
chaotixkilla, line 691: You can refactor all the code related with this to a function I think

```suggestion
    function setManualTrigger (code) {
      if (_.get(code)(this.trader.triggersAutomation) === MANUAL) {
        this.manualTriggersDataProvided[code] = true
      }
    }
```

chaotixkilla (changes requested):

lib/brain.js
```diff
@@ -1949,6 +1973,11 @@ Brain.prototype.smsFlowHandleReturnState = function smsFlowHandleReturnState ()
     return this.runComplianceTiers(nonCompliantTiers)
   }
 
+  // verify if any non compliant manual trigger was scanned and end tx
+  if (this.txBlockedByManualTrigger) {
+    this.txBlockedByManualTrigger = false
```
chaotixkilla, line 1978: Hmm, if I recall correctly, showing the blocked customer screen always leads to calls to `_idle()`. Is it possible to move this `txBlockedByManualTrigger` reset and all its instances (lines 1978, 2204, 2747) to the `_idle()` function instead? It'll just be called once and it's closer to the other variable resets

chaotixkilla (changes requested):

RafaelTaranto (approved):

PR #645, opened 2021-07-21
Insufficient funds screen

chaotixkilla:


chaotixkilla:
[New machine screen for 'insufficient funds' error, in lieu of 'We got this'](https://trello.com/c/wjkyZMSO/839-new-machine-screen-for-insufficient-funds-error-in-lieu-of-we-got-this)

ui/src/app.js
```diff
@@ -198,7 +198,8 @@ function processData (data) {
       break
     case 'balanceLow':
     case 'insufficientFunds':
-      setState('limit_reached')
+      // setState('limit_reached')
```
josepfo, line 201: Unnecessary comment

josepfo (changes requested):
Just a comment. The qr image seems too small compared to the spec and that causes the underlying content to be unaligned with left side. Can you take a look at that?

josepfo (approved):

lib/brain.js
```diff
@@ -3076,10 +3076,11 @@ Brain.prototype._sendCoinsError = function _sendCoinsError (err) {
   const self = this
   if (err.statusCode === INSUFFICIENT_FUNDS_CODE) {
     setTimeout(function () { self._idle() }, self.config.insufficientFundsTimeout)
-    return this._transitionState('insufficientFunds')
+    this._transitionState('insufficientFunds', withdrawFailureRec)
+  } else {
+    this._transitionState('withdrawFailure', withdrawFailureRec)
   }
 
-  this._transitionState('withdrawFailure', withdrawFailureRec)
```
RafaelTaranto, line 3082: Why was this change needed?

RafaelTaranto (changes requested):

lib/brain.js
chaotixkilla: Probably some leftover from earlier development. Fixed

RafaelTaranto (approved):

PR #648, opened 2021-07-22
refactor: identify removable code

josepfo:


josepfo:
[Do not take into account failed/canceled transactions for compliance purposes](https://trello.com/c/bgLnwHCD/886-do-not-take-into-account-failed-canceled-transactions-for-compliance-purposes)

lib/compliance/triggers/triggers.js
```diff
@@ -4,6 +4,8 @@ const BN = require('../../bn')
 const { TRIGGER_TYPES, REQUIREMENTS, DIRECTIONS } = require('./consts')
 
 const filterSinceThreshold = days => _.filter(it => daysSince(it.created) < days)
+// BACKWARDS_COMPATIBILITY 7.5
+// to be deleted: no need to filter as we're doing it on the server-side
```
chaotixkilla, line 8: I think this is enough, as the filter result theoretically will be equal with the starting list of txs, but we can go up a notch and ignore the filtering completely if we want, to save computer resources. Confirm with Rafael if this is something we want

```suggestion
// BACKWARDS_COMPATIBILITY <relevant_version_here>
// to be deleted: no need to filter as we're doing it on the server-side
let filterNoFiatTx
const serverVersion = this.trader.serverVersion
if (!serverVersion || semver.lt(serverVersion, <relevant_version_here>)) {
  filterNoFiatTx = _.filter(it => BN(0).lt(it.fiat || 0))
} else {
  filterNoFiatTx = originalTxs // This line can be done in multiple different ways, find the most elegant one
}
```

chaotixkilla (changes requested):

lib/compliance/triggers/triggers.js
RafaelTaranto: No need for this, complicates the code for low upside.

PR #649, opened 2021-07-23
Add support for individual discounts

chaotixkilla:


chaotixkilla:
[Implement 'Individual discounts', 'Loyalty' panel](https://trello.com/c/suRV3Tf3/889-implement-individual-discounts-loyalty-panel)

josepfo (approved):

josepfo (approved):

lib/brain.js
```diff
@@ -3037,6 +3051,24 @@ Brain.prototype.returnToCashOutState = function returnToCashOutState () {
...
+
+  if (!_.isNil(selectedDiscount)) {
+    /*
+      * It's possible for a customer to insert a promo code before inserting the phone data, leading to two possible discounts.
+      * In that case, use the bigger discount.
+      */
+    const currentDiscount = this.tx.discount || 0
+    if (selectedDiscount.discount > currentDiscount) {
+      this.tx = Tx.update(this.tx, { discount: selectedDiscount.discount, discountSource: 'individualDiscount' })
+    }
```
RafaelTaranto, line 3068: Won't this re-enable the coupon code button in the following scenario?

1. Start the tx with no phone
2. Add a promo code
3. Reache the compliance level of sms verification
4. Have a individual discount bigger than the promo code

lib/brain.js
```diff
@@ -2985,7 +2991,15 @@ Brain.prototype._submitPromoCode = function _submitPromoCode (data) {
...
   this.trader.verifyPromoCode(promoCode, this.tx).then(res => {
-    this.tx = Tx.update(this.tx, { discount: res.promoCode.discount })
+    /*
+      * It's possible for a customer to insert a promo code before inserting the phone data, leading to two possible discounts.
+      * In that case, use the bigger discount.
+      * See: verifyCustomerDiscounts()
+      */
+    const currentDiscount = this.tx.discount || 0
+    if (res.promoCode.discount > currentDiscount) {
+      this.tx = Tx.update(this.tx, { discount: res.promoCode.discount, discountSource: 'promoCode' })
```
RafaelTaranto, line 3001: Instead of having a discount source it might be better to just pass down a flag whenever we use a coupon code. We just update the tx with something like "promoCodeApplied".

If we do that we'll be able to replace this part of the code with the regular `verifyCustomerDiscounts` and fix the bug mentioned on my other comment.

RafaelTaranto (changes requested):

lib/brain.js
chaotixkilla: Not sure if I got the comment correctly (as that bug in specifically does not occur). But I found another interaction during cash-out where this change comes in handy. I'll work on this

RafaelTaranto (approved):

PR #651, opened 2021-08-06
Feat: add delay on tc screen

josepfo:


josepfo:
[Add a delay to acceptance of the Terms & Conditions screen](https://trello.com/c/JGL73LUt/905-add-a-delay-to-acceptance-of-the-terms-conditions-screen)

i18n/ui/lbm-ui_en-US.po
```diff
@@ -491,8 +491,8 @@ msgid "Please press the back of your driver's license against the scan window."
 msgstr "Please press the back of your driver's license against the scan window."
 
 #: On screen: scan-photo-manual
-msgid "Please press the front of your state ID card against the scan window, hold it there, and then press \"Ready to scan\"."
-msgstr "Please press the front of your state ID card against the scan window, hold it there, and then press \"Ready to scan\"."
+msgid "Please press the front of your state ID card against the scan window, hold it there, and then press "Ready to scan"."
+msgstr "Please press the front of your state ID card against the scan window, hold it there, and then press "Ready to scan"."
```
chaotixkilla, line 495: Revert these changes (escaping the `"` char)

ui/src/app.js
```diff
@@ -18,6 +18,9 @@ var isRTL = false
 var two = null
 var cryptomatModel = null
 var termsConditionsTimeout = null
+var termsConditionsAcceptanceInterval = null
+var termsConditionsAcceptanceTimeout = null
+const secondsDelayTermsConditions = 7
```
chaotixkilla, line 23: Doesn't this mean that T&C will always take 7 seconds, independently of what is set on `this.trader.termsDelay`? You are sending the `delay` field on the screen transaction in `brain.js`, but that value is never saved. To do so, add an if statement on the `processData()` function, which is where the data sent from `brain.js` functions is processed

chaotixkilla (changes requested):

RafaelTaranto (approved):

PR #652, opened 2021-08-10
feat: send sms receipt

josepfo:


josepfo:
[Offer SMS receipt](https://trello.com/c/WS8yKiGu/906-offer-sms-receipt)

lib/brain.js
```diff
@@ -2252,7 +2257,7 @@ Brain.prototype._start = function _start () {
 
   this.browser().send({
     tx: this.tx,
-    receiptStatus: this.trader.receiptPrintingActive ? 'available' : 'disabled'
+    receiptStatus: this.trader.receiptPrintingActive ? 'available' : 'disabled',
```
chaotixkilla, line 2260: Any reason for this change? Only thing I could think of is the prettier config with `trailingComma: true`

chaotixkilla: Or, thinking better about this change, seems as a leftover change where you added a field that was later deleted during development

lib/brain.js
```diff
@@ -3369,7 +3410,7 @@ Brain.prototype._chooseFiat = function _chooseFiat () {
 
   this._transitionState('chooseFiat', {
     chooseFiat: response,
-    receiptStatus: this.trader.receiptPrintingActive ? 'available' : 'disabled'
+    receiptStatus: this.trader.receiptPrintingActive ? 'available' : 'disabled',
```
chaotixkilla, line 3413: Same here

lib/trader.js
```diff
@@ -317,6 +317,15 @@ Trader.prototype.triggerSuspend = function triggerSuspend (customerId, triggerId
     .then(r => r.body)
 }
 
+Trader.prototype.smsReceipt = function smsReceipt (data, customerId) {
+  return this.request({
+    path: `/customer/${customerId}/smsreceipt`,
+    body: { data },
```
chaotixkilla, line 323: Isn't `data` already an object?

chaotixkilla: You're accessing this with `req.body.data` on server-side, due to this. Changing this has little to no impact, just pointed it out due to other functions in the trader module sending the object as is, without encapsulating on a `data` field

chaotixkilla (changes requested):

RafaelTaranto (approved):

RafaelTaranto (approved):

PR #660, opened 2021-09-02
feat: pick cam res automatically

josepfo:


josepfo:
[Pick camera resolution automatically](https://trello.com/c/F4r1lEf8/926-pick-camera-resolution-automatically)

chaotixkilla (approved):

lib/scanner.js
```diff
@@ -26,14 +26,30 @@ module.exports = {
...
+const maxCamResolutions = []
+const minCamResolutions = []
+
+function setConfig (cam) {
+  var format = null
+
+  const pixelRes = format => format.width * format.height
+  const sortByPixelRes = _.sortBy(pixelRes)
+  const isSuitableRes = currentRes => {
+    const isUnderMaxAcceptableResolutions = _.map(res => pixelRes(currentRes) < pixelRes(res), maxCamResolutions).every(bool => bool === true)
```
RafaelTaranto: why use `every` here? If we have a list of max resolution we just need to check if it's under one of them, not all.

lib/scanner.js
```diff
@@ -26,14 +26,30 @@ module.exports = {
...
+    const isAboveMinAcceptableResolutions = _.map(res => pixelRes(currentRes) > pixelRes(res), minCamResolutions).every(bool => bool === true)
```
RafaelTaranto: Same as above.

Also a refactor suggestion:

```js
const currentRes = pixelRes(current)

const gteCurrent = _.gt(current)
const lteCurrent = _.gt(current)

const isAboveMinAcceptableResolutions = _.compose(_.some(_.identity), _.map(gteCurrent),_map(pixelRes))
```

lib/scanner.js
```diff
@@ -26,14 +26,30 @@ module.exports = {
...
+    return isUnderMaxAcceptableResolutions && isAboveMinAcceptableResolutions
+  }
+
+  const availableFormats = sortByPixelRes(cam.formats.filter(f => f.formatName === 'MJPG'))
+
+  _.map(elem => {
+    if (isSuitableRes(elem)) format = elem
+  }, availableFormats)
```
RafaelTaranto: How is this picking the best resolution? Isn't it going to be random?

We should get the best possible resolution and cache it, we'll be using that every time so no need to run this again.

RafaelTaranto (changes requested):

lib/scanner.js
josepfo: If I understood this right, we should get a res that doesn't surpass any res value that's on the max list

lib/scanner.js
josepfo: About the questions. We sort the available formats in ascending order, we will get the best suitable res.
But this can be optimized.

PR #661, opened 2021-09-04
feat: paper wallet only from server

josepfo:


josepfo:
[Move paper wallet-only printing to the admin, support the rest of coins](https://trello.com/c/9Li42AQV/934-move-paper-wallet-only-printing-to-the-admin-support-the-rest-of-coins)

chaotixkilla (approved):

lib/brain.js
```diff
@@ -49,6 +49,7 @@ const BigNumber = BN.klass
 
 let transitionTime
 
+const COINS_SUPPORTING_PAPER_WALLET = ['BTC', 'LTC', 'BCH']
```
chaotixkilla: Instead of the hardcoded array here, consider replacing with the `getPaperWalletSupportiveCoins()` function we spoke about

chaotixkilla (changes requested):

chaotixkilla (approved):

PR #670, opened 2021-09-23
chore: update bech32

siiky:


siiky:
[Mock qr code scan failing on dev](https://trello.com/c/M9RPQx16/966-mock-qr-code-scan-failing-on-dev)

PR #677, opened 2021-10-01
fix: non-translated coins

siiky:


siiky:
[Machine not translating coin name](https://trello.com/c/iDltslXJ/984-machine-not-translating-coin-name)

RafaelTaranto (approved):

ui/src/app.js
```diff
@@ -678,7 +678,9 @@ $(document).ready(function () {
       return
     }
 
-    const coin = { cryptoCode: el.data('cryptoCode'), display: el.text() }
+    const cryptoCode = el.data('cryptoCode')
+    const display = currentCoins.find(it => it.cryptoCode === cryptoCode).display
+    const coin = { cryptoCode, display }
     if (!coin.cryptoCode) return
```
joshmh: Should return earlier, by checking ``if (!cryptoCode)``. Also, shouldn't we handle the case where the cryptoCode is not in currentCoins?

joshmh (changes requested):

ui/src/app.js
siiky: > Should return earlier, by checking `if (!cryptoCode)`.

Will do. Just for future reference, when does that happen?

> Also, shouldn't we handle the case where the cryptoCode is not in currentCoins?

I assumed that couldn't happen because the coin comes from the list that comes from the server. But reading again, the `if (!cryptoCode)` also implies it may not come from the list...

joshmh: Thanks! You were checking ``!coin.cryptoCode`` anyway, so might as well not run extra code after finding that it's ``null``. I'm not sure if it can happen but it's still good to check for it because many times things can happen that aren't supposed to happen and these kinds of checks make it easier to find the bug.

siiky: Right, makes sense to me.

PR #709, opened 2021-11-04
feat: optimize disk writes

josepfo:


josepfo:
[Investigate EPIPE errors on ssuboard](https://trello.com/c/jCJtzYGX/1064-investigate-epipe-errors-on-ssuboard)

ubavic (approved):

lib/machine-info.js
```diff
@@ -4,8 +4,12 @@ const writeFile = pify(fs.writeFile)
 const path = require('path')
 const _ = require('lodash/fp')
 
+let MACHINE_INFO = null
```
RafaelTaranto: Uppercase var names are usually reserved to consts.

lib/operator-info.js
```diff
@@ -2,8 +2,13 @@ const pify = require('pify')
 const fs = require('fs')
 const writeFile = pify(fs.writeFile)
 const path = require('path')
+const _ = require('lodash/fp')
+
+let OPERATOR_INFO = null
```
RafaelTaranto: same as above

lib/machine-info.js
RafaelTaranto: Also, since we're keeping this on a local variable, we can load from it as well. We can return early on the load if we already have something on the local variable.

RafaelTaranto (changes requested):

lib/machine-info.js
josepfo: Fair point!

RafaelTaranto (approved):

PR #711, opened 2021-11-15
Fix phone number input

chaotixkilla:


chaotixkilla:
[Fix phone number input on l-m](https://trello.com/c/2e21mNcM/1078-fix-phone-number-input-on-l-m)

PR #722, opened 2021-12-07
fix: show limit reached screen when balance is low

josepfo:


siiky (approved):
LGTM! The `limit_reached` and `out_of_coins` states look reversed to me, but I confirmed with @josepfo that they're right, just misleading.

RafaelTaranto (approved):

PR #733, opened 2022-01-04
Fix network speed precision

chaotixkilla:
fix: increase time scale precision for network speed calculation
feat: use sampling for more accurate download speed values

### DIFF
--- a/lib/ping.js
+++ b/lib/ping.js
@@ -48,10 +48,7 @@ const checkDownloadSpeed = (packages) => {
   }
 
   const promises = _.map(elem => {
-    const sample = []
-    for (let i = 0; i < DOWNLOAD_SAMPLE_SIZE; i++) {
-      sample.push(speedTest)
-    }
+    const sample = _.map(() => speedTest, _.times(_.noop, DOWNLOAD_SAMPLE_SIZE))
 
     const resolveSampleSeq = funcs =>
       funcs.reduce((promise, func) =>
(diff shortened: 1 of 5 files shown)


lib/ping.js
```diff
@@ -19,9 +21,51 @@ const pingRepository = (urls) => {
...
+    const resolveSampleSeq = funcs =>
+      funcs.reduce((promise, func) =>
+        promise.then(result => func(elem.url, elem.size).then(Array.prototype.concat.bind(result))), Promise.resolve([]))
+    
+    return resolveSampleSeq(sample)
+      .then(res => {
+        const speeds = _.map(it => it.mbps * 0.125, res)
+        const sortedAscSpeeds = speeds.sort()
+
+        // Calculate mean after removing top and bottom speed which are possible outliers
```
josepfo, line 62: Can you explain why this can be an issue?

lib/ping.js
```diff
@@ -19,9 +21,51 @@ const pingRepository = (urls) => {
...
+    })
+  }
+
   const promises = _.map(elem => {
-    return testConnection.checkDownloadSpeed(elem.url, elem.size)
-      .then(speed => ({ url: elem.url, speed: speed.mbps * 0.125 }))
+    const sample = []
+    for (let i = 0; i < DOWNLOAD_SAMPLE_SIZE; i++) {
+      sample.push(speedTest)
+    }
```
josepfo: This should be equivalent no?
```suggestion
    const sample = _.times(speedTest, DOWNLOAD_SAMPLE_SIZE)
```

josepfo (changes requested):

josepfo (approved):

PR #738, opened 2022-01-14
Fix manual triggers validation

chaotixkilla:


josepfo (approved):

lib/brain.js
```diff
@@ -2124,18 +2127,19 @@ Brain.prototype.smsFlowHandleReturnState = function smsFlowHandleReturnState ()
   const isCompliant = _.isEmpty(nonCompliantTiers)
   const otherTiers = _.isNil(this.failedCompliance) && !isCompliant
 
+  // verify if any non compliant manual trigger was scanned and end tx
+  if (this.txBlockedByManualTrigger) {
+    this.txBlockedByManualTrigger = false
+    return this.showBlockedCustomer()
+  }
+
```
RafaelTaranto: Isn't moving this here stop at the first manual trigger requested? We don't want to do that.

lib/brain.js
```diff
@@ -2041,13 +2041,16 @@ Brain.prototype.smsCompliance = function smsCompliance (opts = {}) {
...
   const requiredTiers = this.requiredTiers(this.trader.triggers, this.customerTxHistory, triggerTx)
   const requiredManualTiers = _.filter(tier => _.get(tier)(this.trader.triggersAutomation) === MANUAL)(requiredTiers)
   const scannedManualTiers = _.filter(tier => _.get(tier)(this.manualTriggersDataProvided))(requiredManualTiers)
 
+  // If the tier is not manual, allow pending values
   if (!_.includes(tier, requiredManualTiers)) return !_.isEmpty(customerTier)
-  this.txBlockedByManualTrigger = true
+
+  // If the tier is manual, block pending
+  if (!_.isEmpty(customer)) this.txBlockedByManualTrigger = true
```
RafaelTaranto: Why are we testing for customer before this? Won't this skip blocking the tx if a trigger is set as manual to a higher threshold? 

Why even try to pass down customer if you already have the customerTier?

RafaelTaranto (changes requested):

PR #740, opened 2022-01-19
fix: translation

ubavic:


siiky (approved):
LGTM!

ui/src/app.js
```diff
@@ -272,16 +272,24 @@ function processData (data) {
 }
 
 function translate (data, fetchArgs) {
+  if (data === "")
```
RafaelTaranto, line 275: Did you figure out what's causing data to be empty?

ui/src/app.js
```diff
@@ -272,16 +272,24 @@ function processData (data) {
...
-        ? defaultLocale.translate(data).fetch(...fetchArgs)
-        : defaultLocale.translate(data).fetch()
+      try {
+        return fetchArgs
+          ? defaultLocale.translate(data).fetch(...fetchArgs)
+          : defaultLocale.translate(data).fetch()
+      } catch (e) {
+        console.error('Error while translating: ', e)
+        return data
+      }
```
RafaelTaranto, line 292: Also any reason for the default translation to fail? Won't this hide away a issue on the translation itself?

RafaelTaranto (changes requested):

ui/src/app.js
ubavic: There seams that exist a few elements that fail default translation (those element have `baseTranslation` assigned to them, I forgot to fix that). But, same reason as above, I wanted to make `translate` function safer.

ui/src/app.js
ubavic: A lot of elements on custom requirements screens. Those are populated dynamically (after initial translation).

We should (probably) remove i18n tags form those elements (as operator can't translate them), but I put this piece of code just to make `translate` function a little bit safer and faster in case that void tags appear again in future.

ubavic: But, yeah this `if` should be removed.

PR #741, opened 2022-01-20
fix: remove i18n classes for void tags

ubavic:
fix: add baseTranslation translation for buttons


i18n/ui/lbm-ui_en-US.po
```diff
@@ -134,14 +134,6 @@ msgstr "Tap anywhere on the screen when you're done."
 msgid "Network down"
 msgstr "Network down"
 
-#: On screen: custom-permission
-msgid ""
-msgstr ""
-
-#: On screen: custom-permission
-msgid "Continue?"
-msgstr "Continue?"
```
siiky, line 143: > There is one ‘Continue?' button, but we don’t have a translation for that label.

Do you mean in the `i18n/build/*/ui/*.po` files and `locales.js`?

Isn't that actually something that should exist?

siiky (changes requested):

siiky (approved):

PR #750, opened 2022-02-01
fix: use lodash for node6 compat

RafaelTaranto:


package.json
```diff
@@ -35,7 +36,7 @@
     "tar": "~0.1.8",
     "tinycolor2": "^1.4.1",
     "u2f": "^0.1.3",
-    "uuid": "^8.3.2",
+    "uuid": "^3.1.0",
```
siiky, line 39: Do we need both `uuid` and `@fczbkk/uuid4` ?

siiky (approved):

package.json
RafaelTaranto: This is a compatibility fix, I had to add fczbkk version because the uuid version that has `validate` doesn't support node 6

PR #766, opened 2022-02-23
Fix custom info screen

ubavic:


ui/src/app.js
```diff
@@ -331,6 +346,8 @@ function customInfoRequest (customInfoRequest) {
       $('#optional-text-field-2').hide()
       $('.key.backspace.standard-backspace-key').removeClass('backspace-margin-left-override')
       $('.custom-info-request-space-key').show()
+      setComplianceTimeout()
+      customRequirementTextKeyboard.setComplianceTimeout = setComplianceTimeout
```
siiky: Wouldn't it be easier to set this `setComplianceTimeout()` upon instantiation of the keyboards/choice-lists?

```js
customRequirementTextKeyboard = new Keyboard({
  ...,
  setComplianceTimeout: setComplianceTimeout
})
```
and in the constructor
```js
this.setComplianceTimeout = options.setComplianceTimeout
```
Or maybe I'm missing some detail.

siiky (approved):

PR #773, opened 2022-03-16
feat: use GraphQL for all server-machine communication

siiky:
Related to https://github.com/lamassu/lamassu-server/pull/1164


lib/graphql-client.js
```diff
@@ -0,0 +1,63 @@
...
+  new ApolloClient({
+    cache: new InMemoryCache(),
+    defaultOptions: {
+      query: {
+        fetchPolicy: 'no-cache',
+        errorPolicy: 'all',
+      }
+    },
+    link: new HttpLink({
+      uri: 'https://localhost:3000/graphql?' + makeQueryString(opts),
```
chaotixkilla: Shouldn't the `?` only appear if `opts` is not empty or null?

lib/brain.js
```diff
@@ -2043,8 +2042,8 @@ Brain.prototype.smsCompliance = function smsCompliance (opts = {}) {
 }
 
 Brain.prototype.validateTier = function validateTier (tier, triggerTx, customerTier, requiredTiers) {
-  const requiredManualTiers = _.filter(tier => _.get(tier)(this.trader.triggersAutomation) === MANUAL)(requiredTiers)
-  const scannedManualTiers = _.filter(tier => _.get(tier)(this.manualTriggersDataProvided))(requiredManualTiers)
+  const requiredManualTiers = _.filter(tier => !_.get(tier, this.trader.triggersAutomation), requiredTiers)
```
chaotixkilla, line 2037: Were changes regarding this `MANUAL` variable done on the server side? Upon a quick look, didn't find any changes regarding "compensation" to this one, so just asking in case I missed it

lib/trader.js
```diff
@@ -443,7 +680,7 @@ Trader.prototype.pollHandler = function pollHandler (res) {
   this.receiptOptions = _.omit(['active', 'sms'], res.receipt)
   this.enablePaperWalletOnly = res.enablePaperWalletOnly
   this.serverVersion = res.version
-  this.triggersAutomation = res.triggersAutomation
+  this.triggersAutomation = _.mapValues(_.eq('Automatic'), res.triggersAutomation)
```
chaotixkilla, line 715: It's here, alright. Forget the other comment

chaotixkilla (approved):

lib/graphql-client.js
siiky: I think that the query string of an URL can be empty (would have to check to make sure). But the client is only ever used in the trader right now so it won't ever be empty/null. If the client is to be used in the future on more places then this should be changed, yes.

lib/brain.js
siiky: The old poll handler on the server still replies with `Manual` or `Automatic`, for backwards compatibility. But the old reply handler on the machine makes that a boolean, as you saw.

chaotixkilla (approved):

lib/brain.js
```diff
@@ -2652,9 +2651,7 @@ Brain.prototype.sendRates = function sendRates () {
     },
     coins: this.trader.coins,
     twoWayMode: this.twoWayMode(),
-    terms: _.assign(this.trader.terms, {
-      delayTimer: Boolean(this.trader.terms.delay) ? this.config.termsDelay : null
-    }),
+    terms: _.set('delayTimer', this.config.termsDelay, this.trader.terms),
```
RafaelTaranto, line 2646: Aren't we ignoring the "delay" flag by doing this?

RafaelTaranto: Looks like you got your T&C code from pollingRoutes, which is the deprecated version, if you check termsAndConditionsRoutes.js you'll see the delay setting in there.

lib/brain.js
RafaelTaranto: I'm not sure I prefer to have it this way, any reason for the swap from AUTOMATIC|MANUAL to a boolean?

lib/request.js
```diff
@@ -59,4 +59,4 @@ function request (configVersion, globalOptions, options) {
   return got(options.path, gotOptions)
 }
 
-module.exports = {request}
```
RafaelTaranto, line 62: This didn't cause breaking changes?

lib/trader.js
```diff
@@ -231,6 +376,18 @@ function isNetworkError (err) {
...
 }
 
+function isNetworkError (err) {
+  return !!err.networkError
+}
+
+/*
+ * TODO: Remove once the old poller is removed -- replace `this.isNetworkError`
+ * with just `isNetworkError` everywhere!
+ */
```
RafaelTaranto: Can you add the deprecation comment on related places? Check how we did with T&C

lib/trader.js
```diff
@@ -215,12 +356,16 @@ Trader.prototype.postTx = function postTx (tx) {
...
-      if (isNetworkError(err)) return
+      if (this.isNetworkError(err)) return
       throw err
     })
 }
 
-function isNetworkError (err) {
+Trader.prototype.isNewServer = function isNewServer () {
+  return this.serverVersion && semver.gte(this.serverVersion, '8.1.0')
+}
```
RafaelTaranto: This name doesn't tell much, what does new server actually means? Can't we pass down the version as param and have a generic function to check for version?

RafaelTaranto (changes requested):

lib/request.js
siiky: The only other necessary change was a line in `lib/trader.js`: b9b9634105835a2803a15872feec9646178a787a

lib/trader.js
siiky: I did that initially, but it was unnecessarily verbose. It's always called with `this.serverVersion`, so I didn't see that as an improvement.

RE the name: it's a name like any other ^^ I can change it, no problem, but this method is temporary anyway, it won't be necessary after the old poller is removed. And that's what the method means: if the l-s has the GQL server.

lib/brain.js
siiky: This code (the old and the new) just sets `termsDelay` on the `terms` object, which already has the `delay` flag (the resolver code is derived from `termsAndConditionsRoutes.js`).

From the testing I did, the machine behaved as expected according to the `delay` flag. From that I believe (though I didn't check) that the UI code uses the `delay` flag (not `termsDelay`) to decide if there should be a delay.

And I made this change because of an error trying to access `delay` when there were no terms.

lib/trader.js
RafaelTaranto: I mean passing down the required version string as a param, in this case it would be "8.1.0".

`isGraphqlPollEnabled` could be a good one if you don't make it generic.

PR #784, opened 2022-04-06
Fix third fiat button on cash-out transactions

chaotixkilla:


ui/src/app.js
```diff
@@ -1509,14 +1509,12 @@ function getUsedCassettes (cassettes) {
...
-      ? cassettes[2]
-      : doesCassette4HaveBills
-        ? cassettes[3]
-        : cassettes[4]
+  // cassette[2] is the virtual cassette in 2-cassette mode and the third cassette on 3-cassette mode
+  const thirdCassette = 
+    doesCassette4HaveBills
+      ? cassettes[3]
+      : cassettes[2]
   return [cassettes[0], cassettes[1], thirdCassette]
```
siiky: If the (physical) cassettes 3 and 4 of a 4-cassettes machine are empty, doesn't this put the (empty) cassette 4 as the third, when the virtual cassette could be used instead?

If I'm not misunderstanding the problem, wouldn't it be easier to filter and then take (up to) 3 cassettes off the result? `take 3 . filter (\c -> c.count > 0)`, or `_.flow(_.filter(c => c.count > 0), _.take(3))(cassettes)`, or `cassettes.filter(c => c.count > 0).slice(0, 3)` ?

ui/src/app.js
```diff
@@ -1536,8 +1534,7 @@ function setupCassettes (_cassettes) {
       var denomination = formatDenomination(cassette.denomination)
       $('.cash-button[data-denomination-index=' + i + '] .js-denomination').text(denomination)
     }
-    if (doesCassette3HaveBills) $('.cash-button[data-denomination-index=' + 2 + '] .js-denomination').text(formatDenomination(cassettes[2].denomination))
-    if (!doesCassette3HaveBills && doesCassette4HaveBills) $('.cash-button[data-denomination-index=' + 2 + '] .js-denomination').text(formatDenomination(cassettes[3].denomination))
+    $('.cash-button[data-denomination-index=' + 2 + '] .js-denomination').text(formatDenomination(cassettes[doesCassette4HaveBills ? 3 : 2].denomination))
```
siiky: Something similar in this function?

siiky (changes requested):

siiky (approved):

PR #785, opened 2022-04-07
Remove networkDown flag being set to true during a deposit timeout

chaotixkilla:


siiky (changes requested):
I can't even begin to understand this function...

It seems the `networkDown` is only used here:

https://github.com/lamassu/lamassu-machine/blob/687c759787c7714c7971c336b694a6cdffd5a533/lib/brain.js#L3820

But I don't see the point, since, as you pointed out, the trader already emits a `networkDown` when apropriate.

If you're right in removing this line, the `lastGood` lines and the `err.networkDown` I linked to above should be removed as well, they're dead/unnecessary code.

siiky (approved):

PR #790, opened 2022-04-13
feat: cache the T&C based on their hash

siiky:
To be merged after https://github.com/lamassu/lamassu-machine/pull/773. Related: https://github.com/lamassu/lamassu-server/pull/1188.

### DIFF
--- a/lib/trader.js
+++ b/lib/trader.js
@@ -844,6 +844,7 @@ Trader.prototype.oldPollHandler = function oldPollHandler (res) {
 }
 
 Trader.prototype.fetchTerms = function fetchTerms (configVersion) {
+  if (this.isGraphqlPollEnabled()) return
   if (configVersion === this.termsHash) return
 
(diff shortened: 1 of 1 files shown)


lib/trader.js
```diff
@@ -694,8 +698,50 @@ Trader.prototype.pollHandler = function pollHandler (data) {
...
+    if (!terms.details) {
+      // No details even though there's a hash and it differs -- shouldn't happen!
+      this.terms = false
+      this.termsHash = null
+      return
+    }
+
+    this.terms = _.flow(
+      _.pick(['delay', 'title', 'text', 'accept', 'cancel']),
+      _.set('active', true)
```
RafaelTaranto, line 751: Why are we forcing active to true?

RafaelTaranto (changes requested):

lib/trader.js
siiky: The server send the details only if the T&C are active, but it doesn't send the active flag itself because it's implied.

PR #816, opened 2022-07-04
Add automatic receipt printing

chaotixkilla:


lib/trader.js
```diff
@@ -650,9 +650,11 @@ Trader.prototype.pollHandler = function pollHandler (data) {
 
     if (this.receiptPrintingActive = !!data.receiptInfo) {
       this.smsReceiptActive = data.receiptInfo.sms
+      this.automaticReceiptPrintActive = data.receiptInfo.automaticPrint
       this.receiptOptions = _.omit(['sms'], data.receiptInfo)
     } else {
       this.smsReceiptActive = null
+      this.automaticReceiptPrintActive = null
       this.receiptOptions = null
```
siiky: This will conflict with #815.

ui/src/app.js
```diff
@@ -123,6 +125,7 @@ function processData (data) {
   if (data.tx && data.tx.discount) setCurrentDiscount(data.tx.discount)
   if (data.receiptStatus) setReceiptPrint(data.receiptStatus, null)
   if (data.smsReceiptStatus) setReceiptPrint(null, data.smsReceiptStatus)
+  if (data.automaticPrint) setAutomaticPrint(data.automaticPrint)
```
siiky, line 133: Shouldn't `setAutomaticPrint()` be run unconditionally here? There's an if inside on `data.automaticPrint` already.

siiky (approved):
Only one small question, think it looks good.

lib/trader.js
chaotixkilla: Yeah. I'll remind @RafaelTaranto to wait for https://github.com/lamassu/lamassu-machine/pull/815 merge and then I'll rebase this

ui/src/app.js
chaotixkilla: That would make the `else` part be executed everytime, which we do not want

siiky: Hmm I don't really understand it, but you say it's relevant, so be it.

siiky (approved):

PR #820, opened 2022-07-14
Cash collection receipts

chaotixkilla:


josepfo (commented):
Looks good but shouldn't we implement this for zebra printers as well?

josepfo (approved):

PR #821, opened 2022-07-15
Add machine coin rates screen

chaotixkilla:


ui/html/rates.html
```diff
@@ -0,0 +1,26 @@
...
+  <header>
+    <div id="rates-close" class="cancel-button-wrapper">
+      <img class="cancel-button" src="images/close-icon.svg" />
+    </div>
+  </header>
+
+  <main>
+    <div class="rates-header">
+      <h2 class="js-i18n xs-margin-bottom">Exchange Rates</h2>
+      <h3 id="rates-fiat-currency">EUR</h3>
```
siiky, line 16: Shouldn't this be the fiat code configured in the Admin instead of hardcoded to EUR?

ui/src/app.js
```diff
@@ -2136,3 +2144,30 @@ function setReceiptPrint (receiptStatus, smsReceiptStatus) {
       break
   }
 }
+
+function setScreenOptions (opts) {
+  if (opts.rates) {
+    opts.rates.active ? $('#rates-section').show() : $('#rates-section').hide()
+  }
+}
```
siiky: I don't know how/if it could happen, but imagine that `opts.rates.active` is true in an iteration, and on the next iteration there's no `opts.rates`. The rates won't be disabled though they should. Right?

```js
(opts.rates && opts.rates.active) ? $('#rates-section').show() : $('#rates-section').hide()
```

ui/src/app.js
```diff
@@ -2136,3 +2144,30 @@ function setReceiptPrint (receiptStatus, smsReceiptStatus) {
...
+  
+  Object.keys(allRates).forEach(it => {
+    coinEntries.push($(`<div class="xs-margin-bottom">
+    <p class="d2 js-i18n">${allRates[it].cashIn}</p>
+    <h4 class="js-i18n">${it}</h4>
+    <p class="d2 js-i18n">${allRates[it].cashOut}</p>
+  </div>`))
+  })
+
+  $('#rates-fiat-currency').text(fiat)
```
siiky, line 2170: Ah I think you can ignore my other comment about the fiat. This line changes it right?

siiky (changes requested):

ui/src/app.js
chaotixkilla: yes

siiky (approved):

RafaelTaranto:
Needs a rebase + UI rebuild.

PR #835, opened 2022-08-17
Add Fixed Fee setting for cash-out transactions

siiky:


ui/src/app.js
```diff
@@ -111,7 +111,10 @@ function processData (data) {
   if (data.sent && data.total) setPartialSend(data.sent, data.total)
   if (data.readingBill) readingBill(data.readingBill)
   if (data.cryptoCode) translateCoin(data.cryptoCode)
-  if (data.tx && data.tx.cashInFee) setFixedFee(data.tx.cashInFee)
+  if (data.tx) {
+    if (data.tx.cashInFee) setFixedFee(data.tx.cashInFee)
+    else if (data.tx.cashOutFee) setFixedFee(data.tx.cashOutFee)
+  }
```
RafaelTaranto, line 118: This needs to get built, `npm run build`

lib/brain.js
```diff
@@ -3155,16 +3155,7 @@ Brain.prototype.updateBillScreen = function updateBillScreen (blockedCustomer) {
...
-  var billUpdate
-  // BACKWARDS_COMPATIBILITY 7.5.0-beta.1
-  const serverVersion = this.trader.serverVersion
-  if (!serverVersion || semver.lt(serverVersion, '7.5.0-beta.1')) {
-    billUpdate = Tx.billUpdateDeprecated(bill)
-  } else {
-    billUpdate = Tx.billUpdate(bill)
-  }
-
-  return this.fastUpdateTx(billUpdate)
```
RafaelTaranto, line 3174: Yeah, it's time to remove this. But we need to thing of something for this scenario in specific. Maybe throw a error, maybe do it generically on the machine if versions are too far apart. 

@CrypticaScriptura What do you think?

lib/tx.js
```diff
@@ -23,20 +23,22 @@ function mergeTx (oldTx, updateTx) {
...
       _.reduce((acc, v) => acc.add(toCrypto(mergedTx, BN(v.fiat).minus(v.cashInFee))), BN(0), bills),
       cryptoCode),
-    cashInFeeCrypto: truncateCrypto(toCrypto(mergedTx, cashInFee), cryptoCode)
+    cashInFeeCrypto: truncateCrypto(toCrypto(mergedTx, fee), cryptoCode)
   })
 
   const cashOutNewFields = () => ({
     fiat: oldTx.fiat.add(updateTx.fiat || 0),
-    cryptoAtoms: truncateCrypto(toCrypto(mergedTx, oldTx.fiat.add(updateTx.fiat || 0)), cryptoCode)
+    cryptoAtoms: truncateCrypto(toCrypto(mergedTx, oldTx.fiat.add(updateTx.fiat || 0).minus(fee)), cryptoCode),
```
RafaelTaranto: Shouldn't we be adding the fee?

RafaelTaranto (changes requested):

lib/tx.js
siiky: Yeah, nice catch!

lib/brain.js
CrypticaScriptura: Likely time for a version discrepancy screen on the machine that prevents operation entirely if the particular machine's version is sufficiently distant from that of the server. Would be preferable if reserved for major version releases (as this is, v9.0.0).

PR #850, opened 2022-10-06
Add dynamic blacklisted address message

chaotixkilla:


siiky (approved):

siiky (approved):

ui/html/suspicious-address.html
```diff
@@ -10,7 +10,7 @@
   <main>
     <h3 class="js-i18n xs-margin-bottom">Suspicious address</h3>
     <div class="text-content sm-margin-top">
-      <p class="js-i18n d2">
+      <p id="suspicious-address-message" class="js-i18n d2">
         This address may be associated with a deceptive offer or a prohibited group. Please make sure you're using an address from your own wallet.
       </p>
```
RafaelTaranto, line 15: Isn't leaving the message here going to double the "This address..." key on the translations? Might as well remove the text and js-i18n from here.

RafaelTaranto (changes requested):

ui/html/suspicious-address.html
siiky: I didn't see a duplicated message while trying the feature out, but maybe I misunderstood the scenario.

PR #865, opened 2022-11-14
fix: tc scroll display

ubavic:


chaotixkilla (approved):

ui/src/app.js
```diff
@@ -1293,10 +1293,10 @@ function startPage (text, acceptedTerms) {
     textHeightQuantity = document.getElementById('js-terms-text').offsetHeight
     scrollSize = div.offsetHeight - 40
     updateButtonStyles()
-    if (textHeightQuantity <= div.offsetHeight) {
+    if (text.length <= 1000 && textHeightQuantity <= div.offsetHeight) {
       document.getElementById('actions-scroll').style.display = 'none'
     } else {
-      document.getElementById('actions-scroll').style.display = ''
+      document.getElementById('actions-scroll').style.display = 'flex'
```
RafaelTaranto: Oof, we might have a problem here. Any specific reason you chose flex? I don't think the old browsers support this.

RafaelTaranto (changes requested):

ui/src/app.js
ubavic: I think it is best to let this at it was. Property `.style.display` changes only the element's CSS and doesn't modify rule `#actions-scroll`. So, upon setting `...style.display = ''`, flexbox rules from CSS file will be active again. No need to introduce another rule in sass (as we talked).

PR #870, opened 2022-11-18
Implement Sumsub support

chaotixkilla:


lib/brain.js
```diff
@@ -2101,14 +2158,24 @@ Brain.prototype.setManualTrigger = function setManualTrigger (code) {
   }
 }
 
-Brain.prototype.isTierCompliant = function isTierCompliant (tier, triggerTx, requiredTiers) {
+Brain.prototype.isTierCompliant = function isTierCompliant (trigger, triggerTx, requiredTiers, _tier) {
   const tx = this.tx
   const customer = this.customer || {}
 
-  // custom requests have uuid
-  const tierOrCustom = uuid4.validate(tier) ? 'custom' : tier
```
josepfo, line 2130: uuid4.validate() is no longer needed?

lib/compliance/triggers/consts.js
```diff
@@ -19,7 +19,9 @@ const REQUIREMENTS = {
   FACEPHOTO: 'facephoto',
   SUSPEND: 'suspend',
   BLOCK: 'block',
-  US_SSN: 'usSsn'
+  US_SSN: 'usSsn',
+  EXTERNAL: 'external',
+  CUSTOM: 'custom'
 }
 
 const ORDERED_REQUIREMENTS = [
```
josepfo, line 27: Do you know why custom requirements aren't part of this list?

lib/brain.js
```diff
@@ -2454,26 +2527,26 @@ Brain.prototype.isLowBalance = function isLowBalance () {
 }
 
 Brain.prototype.requiredTiers = function requiredTiers (triggers, history, triggerTx) {
-  const triggered = getTriggered(triggers, history, triggerTx)
-  const triggeredFormatted = _.map(o => o.customInfoRequestId ? _.assign(o, { requirement: o.customInfoRequestId }) : o, triggered)
```
josepfo, line 2490: Why did we drop this formatting? Did the triggers object change?

josepfo:
Has conflicts

lib/brain.js
chaotixkilla: Not anymore, UUID was used in a really hacky way to determine if a trigger was a custom info request or not. Basically, if the trigger.tier was a UUID and not a string like `sms` or `usssn`, it would just consider it as a 'custom' tier trigger. We can just use the fields that actually describe a custom info request trigger, instead of this hacky way that was begging for unintended behavior to happen

lib/compliance/triggers/consts.js
chaotixkilla: It was added because 'custom' is now a trigger tier, instead of extrapolating that data based on the validity of a UUID field. Comment above has more context to this change

lib/brain.js
chaotixkilla: See comments above. Most of these changes are to standardize trigger behavior

josepfo (approved):

PR #871, opened 2022-11-21
Replace RxJS

siiky:


lib/brain.js
```diff
@@ -3074,35 +3075,48 @@ Brain.prototype.saveTx = function saveTx (tx) {
...
+      factor: 1.6608,
+      minTimeout: 2000,
+      onFailedAttempt,
     })
+    .then(tx => {
+      this.saveTx(tx) // No big deal if saving the tx details on disk fails
+        .catch(err => console.log("Error saving transaction to local DB (ignoring):", err))
+      return tx
+    })
+    .catch(e => Promise.reject(timedout ? new Error('timeout') : e))
```
chaotixkilla, line 3154: Don't think you need the `Promise.reject` call. It's a catch block, which already returns a Promise

lib/brain.js
```diff
@@ -3074,35 +3075,48 @@ Brain.prototype.saveTx = function saveTx (tx) {
...
+        if (err instanceof E.StaleError || err instanceof E.RatchetError) {
+          console.log(
+            err.name === 'StaleError' ?
+              "Local transaction is outdated, won't retry:" :
+              "Local transaction conflicts with server, won't retry:",
+            err.message
+          )
+          return Promise.reject(new pRetry.AbortError(err))
+        }
+        return Promise.reject(err)
```
chaotixkilla, line 3136: Same here

chaotixkilla (approved):

lib/brain.js
siiky: Yeah, the returned value is casted to a Promise if not a Promise already, but it's resolved, not rejected. So I think it is necessary.

lib/brain.js
siiky: Same here xP

The `pRetry.AbortError` tells `pRetry` not to retry the call in this case.

lib/brain.js
```diff
@@ -3115,6 +3129,10 @@ Brain.prototype.updateTx = function updateTx (updateTx) {
       this.tx = tx
       return tx
     })
+    .catch(e => {
+      console.log(e)
+      return newTx
+    })
```
RafaelTaranto: I'm unsure about this one. There are six functions that call updateTx and only one of them doesn't have proper error handling, with this we'll be masking the error for everyone else.

Also that one usage is in handleScan, at which point there are no bills in the system, so it's better to just fail gracefully over there.

RafaelTaranto (changes requested):

lib/brain.js
siiky: Makes sense, done.

PR #903, opened 2023-01-30
LAM-818 Fix unpairing

chaotixkilla:


lib/trader.js
```diff
@@ -919,7 +930,7 @@ Trader.prototype.pollError = function pollError (err) {
     return
   }
 
-  if (err.statusCode === 403) return this.emit('unpair')
+  if (this.isUnauthorized(err)) return this.emit('unpair')
```
siiky: I think this should be inside the `isNetworkError()` if above. `isUnauthorized()` checks the same variables as `isNetworkError()`, so `isUnauthorized()` is a special case of `isNetworkError()`. Like this, unpair will never be called.

siiky (changes requested):

lib/trader.js
chaotixkilla: You're right, I unintentionally reverted the changes that made it reachable. Fixing it

siiky (approved):

RafaelTaranto (approved):

PR #995, opened 2023-08-09
Draft: printer: add wallet address to wallet printout

dspicher:
It is currently impossible to correlate private key printouts to receipt printouts, because there is no other information besides the private key on the former.

Adding the wallet address to the wallet printout is a natural way of alleviating this.

Draft: This would also need to be done for the `zebra` printer.


dspicher:
@RafaelTaranto we would really benefit from this feature, would it be possible to implement this?

PR #1004, opened 2023-08-22
LAM-101 HCM2

siiky:


lib/brain.js
```diff
@@ -3319,7 +3335,7 @@ Brain.prototype.updateTxCustomerPhoto = function updateTxCustomerPhoto ({ custom
 
 // Don't wait for server response
 Brain.prototype.fastUpdateTx = function fastUpdateTx (updateTx) {
-  console.log('fastUpdateTx', updateTx)
+  console.trace('fastUpdateTx', updateTx)
```
RafaelTaranto, line 3341: We probably want to undo this.

lib/brain.js
```diff
@@ -4206,6 +4222,7 @@ Brain.prototype._dispense = function _dispense () {
 }
 
 Brain.prototype._batchDispense = function _batchDispense (notesToDispense, { batchAmount, dispenseRecords, currentBatch, error }) {
+  console.trace("notesToDispense", notesToDispense)
```
RafaelTaranto, line 4228: Another one to remove

lib/brain.js
```diff
@@ -4243,6 +4260,10 @@ Brain.prototype._batchDispense = function _batchDispense (notesToDispense, { bat
 Brain.prototype._batchesFinished = function _batchesFinished ({ dispenseRecords, error }) {
   const tx = this.tx
   const bills = _.toArray(_.merge(tx.bills, dispenseRecords))
+  console.trace("dispenseRecords:", dispenseRecords)
+  console.trace("error:", error)
+  console.trace("tx:", tx)
+  console.trace("bills:", bills)
```
RafaelTaranto, line 4269: And this

lib/hcm2/hcm2.js
```diff
@@ -9,14 +9,18 @@ const pDelay = require('delay')
 const BN = require('../bn')
 
 const testNotes = 'true'
-const getID = () => ""
+let ID = 0
+const getID = () => {
+  const id = ID++
+  return id.toString()
+}
```
RafaelTaranto, line 16: What happens on a machine restart, is it safe to restart the ID even without restarting the hcm2?

lib/hcm2/hcm2.js
```diff
@@ -28,8 +32,60 @@ const formatTime = datetime =>
...
+    })
+    .then(resolve => ({ resolve }))
+    .catch(reject => ({ reject }))
+  })
+  .then(({ resolve, reject }) => reject ? Promise.reject(reject) : Promise.resolve(resolve))
+  .then(({ body: { error, result } }) => {
+    console.log(path, "error =", error)
+    console.log(path, "result =", result)
+    return error ? Promise.reject(error) : Promise.resolve(result)
+  })
```
RafaelTaranto, line 61: Long form for debugging reasons?

RafaelTaranto (changes requested):

lib/hcm2/hcm2.js
siiky: Yeah, that's not a problem, we could even always send `'0'`. I'm using distinct IDs to associate replies to requests.

lib/hcm2/hcm2.js
siiky: What do you mean? `(...) => { ... }` ?

RafaelTaranto: I don't really get what's the point of this:


```js
    .then(resolve => ({ resolve }))
    .catch(reject => ({ reject }))
  })
  .then(({ resolve, reject }) => reject ? Promise.reject(reject) : Promise.resolve(resolve))
```

siiky: ah that

The function given to the queue must always resolve, no matter what, otherwise it doesn't work properly. But we need/want to tell them apart on the outside. that's all that's doing -- passing the resolve/reject from inside the queue to the outside.

PR #1017, opened 2023-09-14
LAM-914 fix: loose permissions on update folder

RafaelTaranto:


watchdog.js
```diff
@@ -9,7 +9,7 @@ var watchdogInfoLoader = require('./lib/watchdog-info')
 
 require('./lite-logging')
 
-var BASE = '/tmp/extract'
+var BASE = '/opt/lamassu-updates'
 var DONE_PATH = BASE + '/done.txt'
```
siiky: `DONE_PATH` here is `/opt/lamassu-updates/done.txt` but in `lib/update/updater.js` it's `/opt/lamassu-updates/extract/done.txt`

siiky (changes requested):

PR #1039, opened 2023-10-11
LAM-972 refactor: use recyclers intead of stackers

RafaelTaranto:


bin/set-machine-recyclers.js
```diff
@@ -0,0 +1,20 @@
+// Run this after creating the device_config.json file for the Aveiro machines. Only run this on Aveiro machines!
```
siiky: `s/Aveiro/recycler/g`

lib/brain.js
```diff
@@ -2537,9 +2532,9 @@ Brain.prototype._idleTwoWay = function _idleTwoWay (localeInfo) {
   return this.billDispenser.init({
     cassettes,
     fiatCode: this.trader.locale.fiatCode,
-    stackers,
+    recyclers: recyclers,
```
siiky: `s/recyclers: //` just for consistency lol

siiky (changes requested):

hardware/codebase/up4000/aveiro/device_config.json
```diff
@@ -27,7 +27,7 @@
   "billDispenser": {
     "model": "gsr50",
     "cassettes": "1",
-    "stackers": "3"
+    "recyclers": "6"
```
siiky: This field and the above may be numbers, `parseInt(n) = n` when `n` is a number.

PR #1107, opened 2024-03-11
fix: ignore `disable()` before successful startup

siiky:
`initializing` and `initialized` are now common to the validator and dispenser. `disable()` calls `openCloseShutter()` only if the startup has been successful.

### DIFF
--- a/lib/hcm2/hcm2.js
+++ b/lib/hcm2/hcm2.js
@@ -9,7 +9,7 @@ const pDelay = require('delay')
 const BN = require('../bn')
 
 const testNotes = 'false'
-const LDN = 'HCM2_01'
+let LDN = 'HCM2_01'
 
 let ID = 0
 const getID = () => {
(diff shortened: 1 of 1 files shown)


lib/hcm2/hcm2.js
```diff
@@ -9,7 +9,7 @@ const pDelay = require('delay')
 const BN = require('../bn')
 
 const testNotes = 'false'
-const LDN = 'HCM2_01'
+let LDN = 'HCM2_01'
```
RafaelTaranto, line 12: We can lowercase this if it's not a const anymore

RafaelTaranto (changes requested):

lib/hcm2/hcm2.js
siiky: I'd have to change all other lines using it though:

```js
{
  LDN: ldn
}
```

Instead of

```
{
  LDN
}
```

The only reason it was uppercase to begin with.

PR #1125, opened 2024-04-30
chore: change `lamassu-browser` user to `ubilinux`

siiky:


hardware/system/ssuboard/supervisor/conf.d/lamassu-browser.conf
```diff
@@ -1,7 +1,7 @@
 [program:lamassu-browser]
 command=/usr/local/bin/surf -KF /opt/lamassu-machine/ui/start.html
 environment=DISPLAY=":0"
-user=debian
```
CrypticaScriptura: We should keep this as `debian` for the `ssuboard` platform.

PR #1134, opened 2024-05-28
LAM-1103 feat: camera diagnostics

RafaelTaranto:


lib/scanner.js
```diff
@@ -272,6 +276,22 @@ const delayedPhoto = (delay, callback) => {
     .catch(err => callback(err, null))
 }
 
+const diagnosticPhotos = () => {
+  cameraStreamer.setVerbose(true)
+  const promises = [
+    delayedshot('', 1, '/dev/video-scan'),
+    delayedshot('', 1, '/dev/video-front')
```
siiky, line 283: This will not work correctly in those machines (CC?) that use the single camera as both scan and front. The two calls, if scan is the same as front, must be sequential.

RafaelTaranto: I don't think that's the issue since we don't mount the same camera into two mount points, simply have a different mount point to feature mapping for some machines. Probably something to do with the camera-streamer code on the lamassu-machine side, will serialize it anyways.

PR #1152, opened 2024-07-18
LAM-1128 feat: make diagnostics delay configurable through `device_config.json`

siiky:


lib/scanner.js
```diff
@@ -14,6 +14,8 @@ cameraStreamer.setVerbose(IS_VERBOSE)
 let configuration = null
 let kogoroshiya = null
 let DEFAULT_FPS = 10
+const DEFAULT_DELAYEDSHOT_DELAY = 3
+let DELAYEDSHOT_DELAY = DEFAULT_DELAYEDSHOT_DELAY
```
RafaelTaranto: Full uppercase is for consts, make this a regular variable with snake case.

RafaelTaranto (changes requested):
You also need to adapt the other scanner files. I think we have mock and genmega scanners that need to follow standard.

PR #1186, opened 2024-09-13
LAM-1166 Chore/remove deprecated code

RafaelTaranto:


lib/tx.js
siiky, line 1: Can't `billUpdateDeprecated()` and `createBillDeprecated()` be removed?

siiky (changes requested):

siiky (approved):

PR #1190, opened 2024-09-20
LAM-1169 feat: cache configs to disk

siiky:


lib/variable-interval.js
```diff
@@ -24,11 +24,8 @@ const setVariableInterval = function (_thunk, interval, nextInterval) {
 }
 
 const clearVariableInterval = vi => {
-  if (!vi) return
-  if (vi.handle) {
-    clearInterval(vi.handle)
-    vi.handle = null
-  }
+  if (vi?.handle) vi.handle = clearInterval(vi.handle)
+  return null
```
RafaelTaranto: Not sure I understand these changes. 

1. clearInterval returns undefined and so you can reuse it, but it's kinda implicit and I'm sure we'll have to see docs for clearInterval whenever we're looking at this code. Better to just explicitly assign null. 

2. Why was the return added? We don't really use the return of this function.

lib/sqlite.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+const path = require('path')
+const { open } = require('sqlite')
+const sqlite3 = require('sqlite3')
+if (false) sqlite3.verbose()
```
RafaelTaranto: leftovers?

lib/sqlite.js
```diff
@@ -0,0 +1,51 @@
...
+
+const mode = readwrite =>
+  readwrite ?
+    sqlite3.OPEN_READWRITE | sqlite3.OPEN_CREATE :
+    sqlite3.OPEN_READONLY // TODO: check if opening fails
```
RafaelTaranto: What's this TODO? Any reason we're not solving it before pushing to dev?

lib/db.js
```diff
@@ -12,24 +11,24 @@ let dbName
 module.exports = { save, prune, clean }
 
 function list (dbRoot) {
-  return fs.mkdir(dbRoot)
+  return fs.promises.mkdir(dbRoot)
```
RafaelTaranto: This applies for the whole file. While fs.promises still works it's no longer the official example in the docs, might as well migrate to the require flavor of `('node:fs/promises')`. 

You can also destruct the import to have only the used functions. This doesn't really apply for this example but it's good practice for tree shaking.

With the proposed changes future api breakage for node would imply in only changing the require line instead of adjusting it throughout the code.

lib/sqlite.js
```diff
@@ -0,0 +1,51 @@
+'use strict'
+const path = require('path')
+const { open } = require('sqlite')
```
RafaelTaranto: What's the advantage of using open from sqlite instead of doing `new sqlite3.Database()`? Is it worth the added dependency?

lib/sqlite.js
```diff
@@ -0,0 +1,51 @@
...
+    sqlite3.OPEN_READONLY // TODO: check if opening fails
+
+const openWithDriver = driver => (filename, readwrite=false) =>
+  open({
+    filename,
+    mode: mode(readwrite),
+    driver,
+  })
+
+exports.open = openWithDriver(sqlite3.Database)
```
RafaelTaranto: Don't see this being used.

lib/sqlite.js
```diff
@@ -0,0 +1,51 @@
...
+  const all = exports.all(withDB)
+  const get = exports.get(withDB)
+  const run = exports.run(withDB)
+  const withTx = exports.withTx(withDB)
+  return { withDB, all, get, run, withTx }
+}
+
+const repeat = (n, x) => Array(n).fill(x)
+exports.makeTuple = n => '(' + repeat(n, '?').join(',') + ')'
+exports.makeValues = (nColumns, nRows) => repeat(nRows, exports.makeTuple(nColumns)).join(',')
```
RafaelTaranto: I'm confused with all of the exports, most of them are not really being used anywhere and it makes an awkward read. 

I'm also failing to understand the benefit of everything being 'factory' based when it uses the exact same db object anyways. Why not just do `db.saveCoins` (or whatever queries) on lssettings?

RafaelTaranto (changes requested):

RafaelTaranto (changes requested):
Another more general thing. This would benefit from having a separate folder under lib, either `/db`, `/settings`, `/config` or something in this vein. We can then have a pretty closed module. 

Would be good to add testing for this as well, you can use `vitest`, maybe will have to search on how to use vitest without vite.

lib/sqlite.js
siiky: `sqlite` is a thin wrapper for `sqlite3`-like libraries. The API is [promise-based](https://github.com/kriasoft/node-sqlite#usage) instead of [callback-based](https://github.com/TryGhost/node-sqlite3#usage), and adds the basic [migrations mechanism](https://github.com/kriasoft/node-sqlite#migrations).

lib/variable-interval.js
siiky: 1. Done.
2. To make the falsy return explicit. But the intent of this module's interface is to be as similar to `setInterval()`/`clearInterval()` as possible, so I removed the explicit return.

lib/sqlite.js
siiky: `all`/`get`/`run` were used at some point, until I switched to `withTx`.

siiky: RE factory based: if you mean the `helpers()` it's for usage-convenience.

lib/sqlite.js
siiky: I forgot about this one... Just confirmed that opening in readonly mode fails if the file doesn't exist. This will be a bit of a pain to work around...

RafaelTaranto (approved):

PR #1226, opened 2024-11-18
LAM-1193 Feat/nodejs camera

RafaelTaranto:


lib/scanner-node.js
```diff
@@ -0,0 +1,267 @@
...
+const DEFAULT_FPS = 10
+const DEFAULT_DELAYEDSHOT_DELAY = 3
+
+let configuration = null
+let current_fps = DEFAULT_FPS
+let delayedshot_delay = DEFAULT_DELAYEDSHOT_DELAY
+
+let activeStream = null
+
+const maxCamResolutions = [
```
siiky, line 18: `maxCamResolutions`, `maxCamResolutions`, and `maxCamResolutionQRCode` are defined in `consts.js`. Maybe `maxCamResolutionPhotoId` could be added there too, though it might currently not be used anywhere else.

lib/brain.js
```diff
@@ -2564,8 +2540,8 @@ Brain.prototype._startAddressScan = function _startAddressScan () {
 
   const frames = []
   const saveFailedScans = this.shouldSaveQrAttempts ?
-    readFailedFrames(photo => frames.push(photo)) :
-    null
+    photo => frames.push(photo) :
+    _.noop
```
siiky, line 2544: There's no limit to the number of frames here, and `lib/scanner-node.js:capture()` calls it on every single failed frame (at most 1 frame/s), which will surely fail to upload in `Brain.failedQRScans()`.

lib/brain.js
```diff
@@ -743,35 +743,11 @@ Brain.prototype.timeoutToScannerCancel = function timeoutToScannerCancel (stateM
...
-  .catch(console.error)
-
 Brain.prototype.scanPDF = function scanPDF () {
   const customer = this.customer
   // TODO: Change the LED code to support genmega
   this.scanBayLightOn()
-  const saveFailedScans = this.shouldSaveQrAttempts ?
-    readFailedFrames(photo => idCardData.addIdDataPhoto(photo)) :
-    null
+  const idCardStillsCallback = photo => idCardData.addIdDataPhoto(photo)
```
siiky, line 750: This will use the first 5 failed frames (which could be the first 5s of a 30s-60s interval), which is not necessarily a good sample.

lib/capture/index.js
siiky, line 1: This file doesn't appear to be used?

lib/capture/index.js
```diff
@@ -0,0 +1,86 @@
...
+}
+
+const scanPDF417 = (callback, idCardStillsCallback) => {
+  const mode = 'photoId'
+  const device = getCameraDevice(mode)
+
+  let processing = false
+
+  const camStream = stream(device, {})
+  activeStreams.set(device, camStream)
```
siiky, line 46: What if a stream already exists for this `device`?

lib/capture/scanner/genmega.js
siiky, line 1: This file also doesn't appear to be used?

lib/capture/streamer/mock.js
```diff
@@ -0,0 +1,81 @@
...
+      const { width, height } = imageMetadata
+      readableStream.push({ frame: cachedFrame, width, height })
+    }
+
+    if (isStreaming) {
+      setTimeout(sendFrame, 100) // Simulate camera frame rate
+    }
+  }
+
+  initialize()
```
siiky, line 56: `initialize()` is `async` but not `await`ed for.

lib/capture/streamer/mock.js
```diff
@@ -0,0 +1,81 @@
...
+
+  const timeoutHandler = setTimeout(() => {
+    readableStream.destroy()
```
siiky, line 59: In this case `isStreaming` remains `true`, right?

lib/capture/streamer/mock.js
```diff
@@ -0,0 +1,81 @@
...
+  }
+
+  readableStream.on('close', cleanup)
+  readableStream.on('end', cleanup)
+
+  return readableStream
+}
+
+const cameraExists = () => {
+  const mockImagePath = path.join(IMAGE_PATH)
```
siiky, line 74: :monocle_face:

lib/capture/streamer/mock.js
```diff
@@ -0,0 +1,81 @@
...
+  
+  const readableStream = new Readable({
+    objectMode: true,
+    read() {}
+  })
+
+  // Mock image path
+  const mockImagePath = path.join(IMAGE_PATH)
+  
+  if (!fs.existsSync(mockImagePath)) {
```
siiky, line 20: Probably unnecessary, an error would be thrown later in `initialize()` if the file doesn't exist anyway (plus, it's a classic check before use race condition).

lib/capture/streamer/v4l2camera.js
```diff
@@ -0,0 +1,86 @@
...
+        .finally(() => {
+          setTimeout(captureFrame)
+        })
+    })
+  }
+
+  captureFrame()
+
+  const timeoutHandler = setTimeout(() => {
+    readableStream.destroy()
```
siiky, line 59: Same as in the mock streamer: `isStreaming` remains true, right?

lib/capture/streamer/mock.js
RafaelTaranto: Destroy emits a 'close' event.

lib/capture/streamer/mock.js
RafaelTaranto: This is because I don't want to make the interface async, other streams just return. Also since we return the stream and then once initialize is done we start sending frames, it doesn't really impact how the mock works.

PR #1262, opened 2025-01-05
LAM-1279 LAM-1275 refactor: use knex for db

RafaelTaranto:


lib/storage/settings/db.js
```diff
@@ -0,0 +1,42 @@
...
+    afterCreate: (conn, cb) => conn.run('PRAGMA foreign_keys = ON', cb)
+  },
+  ...knexStringcase.default({
+    appPostProcessResponse: (result, queryContext) => {
+      if (queryContext?.booleanColumns) {
+        const processBooleans = (row) => {
+          const processed = { ...row }
+          queryContext.booleanColumns.forEach(col => {
+            if (col in processed && (processed[col] === 1 || processed[col] === 0)) {
+              processed[col] = Boolean(processed[col])
```
siiky, line 24: Got a few questions about this because I'm not seeing the purpose...

1. Why is the `row` spread in a new object, `processed`?
2. Why does it confirm `col` is in `processed`? If `col` is in the `booleanColumns`, isn't it expected to be in `processed`?
3. Why is the value of `col` confirmed to be `1` or `0`? Because of `NULL`s?

lib/storage/settings/settings.fixtures.js
```diff
@@ -0,0 +1,143 @@
+const timezone = new Date()
```
siiky, line 1: [`timezone`](https://github.com/lamassu/lamassu-server/blob/v10.0.7/lib/graphql/types.js#L141) is an UTC offset (-12...+14), not a date or timestamp.

lib/storage/settings/migrations/20241207173401_migration_name.js
```diff
@@ -0,0 +1,165 @@
...
+*
+*/
+exports.up = function (knex) {
+  return knex.schema
+    .createTable('staticConfig', function (table) {
+      table.integer('version').notNullable()
+      table.boolean('enablePaperWalletOnly').notNullable()
+      table.boolean('hasLightning').notNullable()
+      table.text('serverVersion').notNullable()
+      table.timestamp('timezone').notNullable()
```
siiky, line 13: [`timezone`](https://github.com/lamassu/lamassu-server/blob/v10.0.7/lib/graphql/types.js#L141) is an UTC offset (-12...+14), not a date or timestamp.

lib/trader.js
```diff
@@ -956,43 +956,38 @@ Trader.prototype.failedQRScans = function failedQRScans (frames) {
...
     }
   }
 
-  const loadTerms = ([hash, terms]) => {
-    this.termsHash = hash
-    this.terms = terms
-  }
-
-  return lssettings.loadConfig()
+  return settings.loadConfig()
```
siiky, line 987: `machineInfo` is missing. It appears to be missing already from when I worked on it... `loadConfig()` has to be changed, and the related tests.

lib/storage/settings/settings.test.js
```diff
@@ -0,0 +1,238 @@
...
+      random: false,
+      field: true,
+      newOption: true
+    }
+
+    await saveConfig({ ...save, receiptOptions: newReceiptOptions })
+    const { receiptOptions } = await loadConfig()
+    expect(receiptOptions.random).toBe(false)
+    expect(receiptOptions.field).toBe(true)
+    expect(receiptOptions.newOption).toBe(true)
```
siiky, line 213: I think this one could be stronger, like the one below? `.toEqual(newReceiptOptions)`

lib/storage/settings/settings.js
```diff
@@ -0,0 +1,307 @@
...
+    .leftJoin('customRequests', 'customInfoRequests.customRequest', 'customRequests.rowid')
+    .leftJoin('customInputs', 'customRequests.input', 'customInputs.rowid')
+    .leftJoin('customScreen as screen1', 'customRequests.screen1', 'screen1.rowid')
+    .leftJoin('customScreen as screen2', 'customRequests.screen2', 'screen2.rowid')
+    .queryContext({ booleanColumns: ['cirEnabled'] })
+    .then(async rows => {
+      // Get all choice lists in one query
+      const choiceLists = await trx('customInputChoiceList')
+        .select(['customInput', 'choiceText'])
+        .orderBy('rowid')
```
siiky, line 40: Can't this be executed concurrently with `trx('triggers')`? Smth like:

```js
return Promise.all([
  trx('triggers')...,
  trx('customInputChoiceList')...,
]).then(([triggers, choiceLists]) => ...)
```

lib/storage/settings/settings.js
```diff
@@ -0,0 +1,307 @@
...
+      }).returning('rowid')
+
+      return trx('customInfoRequests').insert({
+        id: customInfoRequest.id,
+        enabled: customInfoRequest.enabled,
+        customRequest: customRequestDb.rowid
+      }).returning('id')
+    }
+
+    const saveTriggers = trx('triggers').del()
```
siiky, line 240: A bit picky, but took me a while to find this `.del()`... Maybe it could go on its own line?

PR #1288, opened 2025-02-21
LAM-1193 PR fixes for node camera

RafaelTaranto:


lib/brain.js
```diff
@@ -2455,7 +2455,7 @@ Brain.prototype._privateWalletPrinting = function _privateWalletPrinting () {
 
 Brain.prototype.failedQRScans = function failedQRScans (frames) {
   return this.shouldSaveQrAttempts ?
-    this.trader.failedQRScans(frames) :
+    this.trader.failedQRScans(_.take(5, frames)) :
```
siiky: The problem with this is that only the first 5 photos will be used, which correspond to the first \~5s of 30\~60s. It'd be better to chunk all failed scans and take 5 of them at roughly evenly spaced intervals, [as before](https://github.com/lamassu/lamassu-machine/commit/473c97722af3dad7424af9ea366a88d97f46fce3#diff-7077a8448a960187c5eaa685e0c580c806e896534d9608390b690f6f55cca0ccL830).

siiky (changes requested):

siiky (approved):

PR #1359, opened 2025-06-18
LAM-1137 fix: wait for bill validator before going to idle

siiky:


lib/brain.js
```diff
@@ -2346,7 +2346,15 @@ Brain.prototype._idleTwoWay = function _idleTwoWay (localeInfo) {
...
-  this._chooseCoinScreen(localeInfo)
+  new Promise((resolve, reject) => {
+    const isValidatorEnabledInterval = setInterval(() => {
+      if (this.billValidatorEnabled) {
+        clearInterval(isValidatorEnabledInterval)
+        resolve()
+      }
+    }, 100)
+  })
+    .then(() => this._chooseCoinScreen(localeInfo))
```
RafaelTaranto: I don't think you should retry this, it should be like `idleTwoWay` since `_idle` is already retrying. If you leave it like this you're going to trigger `_chooseCoinScreen` multilpe times

RafaelTaranto (changes requested):

RafaelTaranto (approved):

PR #1376, opened 2025-08-20
LAM-1383 refactor: make liveview self-contained

siiky:


lib/brain.js
```diff
@@ -122,15 +122,23 @@ const Brain = function (config) {
...
 
   this.shouldSaveQrAttempts = config.qrSnapshotDebug
-  this.scanner = config.mockCam
-    ? require('./mocks/scanner')
-    : require(`./${this.isGenmegaMachine ? 'scanner-genmega' : 'scanner-node'}`)
+  this.scanner = require(
+    config.mockCam ? './mocks/scanner' :
+    this.cryptomatMaker === 'genmega' ? './scanner-genmega' :
+    './scanner-node'
+  )
```
RafaelTaranto: This became significantly harder to parse. Might as well just leave as it was.

lib/brain.js
```diff
@@ -122,15 +122,23 @@ const Brain = function (config) {
...
 
-  this.scanner.config(config)
+  const scannerConfig = {
+    ...config,
+    liveviewEnabled: ['coincloud'].includes(this.cryptomatMaker)
+      && !['genmega'].includes(this.cryptomatMaker)
+  }
+  this.scanner.config(scannerConfig)
+  this.liveviewEnabled = scannerConfig.liveviewEnabled // scanners can force-disable liveview
```
RafaelTaranto: Is there a reason to have a list of disables and the force disable on each scanner? Might as well just make sure force disable all non camera scanners.

Also can you move the liveviewEnabled thing to a boolean declaration outside the object? It helps with reading.

RafaelTaranto (changes requested):

PR #1399, opened 2025-11-28
LAM-1506 refactor: remove lamassu coins

RafaelTaranto:


lib/coin-config.js
siiky, line 1: if coin configs aren't saved to disk then `storage/` doesn't seem like the right place (maybe unless we intend to save them to disk in the future)

lib/storage/coin-config.js
```diff
@@ -0,0 +1,40 @@
...
+const getZeroConf = (cryptoCode) => getCoinConfig(cryptoCode).zeroConf ?? false
+
+const getUnitScale = (cryptoCode) => {
+  const coinConfig = getCoinConfig(cryptoCode)
+  if (!_.isNumber(coinConfig.unitScale)) throw new Error(`unitScale is not a number for ${cryptoCode}`)
+
+  return coinConfig.unitScale
+}
+
+const getDepositUrl = (cryptoCode, address, amount) => {
```
siiky: this function also feels a bit out of place...

lib/scanner-node.js
```diff
@@ -195,15 +193,10 @@ const scanQR = ({ resultCallback }) => {
...
+const scanMainQR = ({ resultCallback, stillsCallback }) => {
   const mode = 'qr'
   const device = getCameraDevice(mode)
-
-  const processCallback = async ({ frame, width, height }) => {
-    const code = await qrScanner.scanQRcode({ frame, width, height })
-    if (!code) return null
-    return coinUtils.parseUrl(cryptoCode, NETWORK, code)
-  }
+  const processCallback = qrScanner.scanQRcode
```
siiky, line 188: Shouldn't it at least check the address is valid? Why leave that for the `brain.js`?

Though this way `scanQR` and `scanMainQR` can be merged...

lib/scanner-genmega.js
```diff
@@ -66,11 +63,7 @@ const scanMainQR = ({ resultCallback, cryptoCode }) => {
         return resultCallback(null, null)
       }
       console.log('DEBUG55: %s', decoded)
-      try {
-        resultCallback(null, coinUtils.parseUrl(cryptoCode, NETWORK, decoded))
-      } catch (error) {
-        resultCallback(error)
-      }
+      resultCallback(null, decoded)
```
siiky, line 55: Same as `lib/scanner-node.js`

lib/crypto/ln-utils.js
```diff
@@ -0,0 +1,54 @@
...
+  const isInvoice = hasValidPrefix(address, network, invoiceOptions)
+  const isLnurl = hasValidPrefix(address, network, lnurlOptions)
+
+  try {
+    isInvoice ? bolt11.decode(address) : bech32.decode(address, Number.MAX_SAFE_INTEGER)
+  } catch (error) {
+    return false
+  }
+
+  return isInvoice || isLnurl
```
siiky, line 31: Calling `bech32.decode()` with possibly bad data causes some urticária... Avoiding it (such as with an early return before the `try` block) would be best.

```js
if (!isInvoice && !isLnurl)
  return false
```

siiky (changes requested):

lib/scanner-node.js
RafaelTaranto: Right, my thinking here is that it's not responsibility of a scanning module to parse the result of the scan. We don't really do auto retries anyways since we have interrupting UI. 

Scanners would scan, and whatever needs to parse would parse. Brain.js would orchestrate things, I think this is what it should only be doing in the first place.

siiky (approved):

